PR #108, opened 2015-05-31
Pebble support

looseend:
This should send the live ride notifications to the pebble if exists and the app is installed.

If there is no watch connected it won't try sending the notifications.


jezhiggins:
Hi Jon, sorry for taking to long to press the button on this.

Issue #68, opened 2012-09-06
Font size - too small


si-the-pie:
We've had this [feedback #6585](http://www.cyclestreets.net/views/feedback/6585/):

"The android app is lovely and smooth but it would be so much more usable if only the font used were larger. On my Galaxy Nexus, which has a large screen for a smartphone, street names and similar labels resolve to a font of about 6 points height at high zoom values which is more or less illegible. Certainly I couldn't read it if the phone were clamped to my handlebars."

I've also had one other complain verbally to me about this small size recently.

yk4ever:
It's almost April now, and street names are still unreadable. This should be reopened.

mvl22:
@yk4ever The latest release just went live - do give it a try. This adds support for high resolution screens.

https://play.google.com/store/apps/details?id=net.cyclestreets

mvl22:
We are now serving native 2x tiles to the app for the OpenCycleMap style, as of
https://github.com/cyclestreets/tilecache/commit/6469ae1bebe04f48bba567d34d9a955321d16f9d

They now look much nicer - the 4 pixels to one dot pixellation is now gone.

KonstantinYegupov:
Stuff still looks pixelated on my 1080p device, although the street names are at least legible now.

PR #112, opened 2015-08-23
Switch map tiles to smooth upscaling 

kyegupov:
On high-density screens (https://github.com/cyclestreets/android/issues/68), render map tiles with smooth upscaling, rather than nearest neighbor.


oliverlockwood:
@jezhiggins @mvl22 as this PR is over a year old and not mergeable, should we close it for now?  It can always be referred back to or re-opened if necessary.

Issue #128, opened 2016-05-14
Blog loop


dave9011:
Several users reported an endless loop when opening the blog. The app opens the blog in the default browser and, when the user attempts to press back to return to app, the browser instead reloads the blog.

PR #129, opened 2016-05-14
Update blog URL scheme from HTTP to HTTPS. This allows the app to ope…

dave9011:
Attempt to solve issue #128 . Updated the blog URL scheme from HTTP to HTTPS. This allows the app to open the blog within its webview rather than the browser.

### DIFF
--- a/cyclestreets.vNext/src/main/java/net/cyclestreets/BlogFragment.java
+++ b/cyclestreets.vNext/src/main/java/net/cyclestreets/BlogFragment.java
@@ -6,7 +6,6 @@ import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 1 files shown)


jezhiggins:
Thanks

mvl22:
Can you clarify what this commit does?

We should be using https:// for it now, so not sure why there is an http:// reference there still.

jezhiggins:
On 15/05/16 12:13, Martin wrote:
> Can you clarify what this commit does?
> 
> We should be using https:// for it now, so not sure why there is an
> http:// reference there still.
> 

There was a mistake in the pull request, which I corrected -

https://github.com/cyclestreets/android/commit/0dd297a644418c53291a2b6d953879d4bb9f9251

mvl22:
Thanks!

PR #132, opened 2016-05-27
Upgrade minor/patch versions of dependencies

oliverlockwood:
-  Upgrade minor/patch versions of dependencies
-  While we're at it, fix alignment and conform with convention (as I understand it!) for Gradle dependency definition layout.

Note that `org.osmdroid:osmdroid-android:4.3' has a major upgrade available of `5.2`.  I haven't changed that in this PR as there are associated API changes.

### DIFF
--- a/libraries/cyclestreets-core/build.gradle
+++ b/libraries/cyclestreets-core/build.gradle
@@ -1,13 +1,11 @@
-
-
 configurations {
   compile.exclude module: 'commons-logging'
+  compile.exclude module: 'httpclient'
 }
 
 dependencies {
-  compile 'org.osmdroid:osmdroid-android:4.3',
-      'org.apache.james:apache-mime4j:0.6',
-      'org.apache.httpcomponents:httpmime:4.1',
-      'org.slf4j:slf4j-android:1.6.1-RC1'
+  compile 'org.osmdroid:osmdroid-android:4.3'
+  compile 'org.apache.james:apache-mime4j:0.7.2'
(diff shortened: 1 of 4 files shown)


libraries/cyclestreets-core/build.gradle
```diff
@@ -1,13 +1,11 @@
-
-
 configurations {
   compile.exclude module: 'commons-logging'
+  compile.exclude module: 'httpclient'
```
oliverlockwood: This exclusion is necessary as it seems the latest version of `httpmime` pulls in `httpcomponents` transitively, which clashes with what Android includes natively.
>WARNING: Dependency org.apache.httpcomponents:httpclient:4.5.2 is ignored for <build target> as it may be conflicting with the internal version provided by Android.
         In case of problem, please repackage it with jarjar to change the class packages

mvl22:
Keeping the app at latest OSMdroid in general we should aim for - I get the general feeling they're working on lots of improvements, and the current map framework doesn't feel as smooth/polished as the iOS app has.

oliverlockwood:
@mvl22 good to know, I'll look at raising another PR to update that as a followup.

oliverlockwood:
@jezhiggins ping - I'm holding off the OSMdroid upgrade work until you're happy with the existing PRs.

jezhiggins:
> @jezhiggins ping - I'm holding off the OSMdroid upgrade work until you're
> happy with the existing PRs.

Hi - thanks for the prod.  Delay is due to inaction rather than
unhappiness on my part.  I shall give myself a kick up the arse.

Issue #133, opened 2016-06-10
Upgrading osmdroid dependency to 5.x


oliverlockwood:
This issue is to track upgrading the `osmdroid` dependency, which I've started looking at.

There's quite a lot of changes after 4.3, so my thought was to work through iteratively by first upgrading to 5.0.1, resolving issues, and then moving on to 5.1 and 5.2.  From experimentation:

#### Upgrading to 5.0.1
In this version, the main non-backward-compatible change that affects us at compile time appears to be the removal of the `org.osmdroid.http` package, which was done in [this commit](https://github.com/osmdroid/osmdroid/commit/7f6191f57593f46fd5a903c6fe7d86aadf6d7db2).  From the commit message, it appears to be a desire to move away from Apache HttpClient towards  UrlConnection.
I can see more than one option on how to proceed:
1.  Continue using Apache HttpClient and implement our own form of the `HttpClientFactory` classes that have been removed from `osmdroid`.
2.  Migrate our code that makes REST requests to use UrlConnection instead.

From reading around blogs, StackOverflow etc I don't think (1) is the way forward, as it seems that Google is not really supporting Apache HttpClient going forward.  Quote from http://android-developers.blogspot.co.uk/2011/09/androids-http-clients.html:
>Apache HTTP client has fewer bugs on Eclair and Froyo. It is the best choice for these releases.
>For Gingerbread and better, HttpURLConnection is the best choice.
>New applications should use HttpURLConnection; it is where we will be spending our energy going forward.

Since our minSdkVersion is already 14 (Ice Cream Sandwich) it seems to make sense for us to go down the UrlConnection route, but I'd be grateful for opinions from the maintainers. @jezhiggins @mvl22, what are your thoughts on the above - and, indeed, do you have any better ideas?

oliverlockwood:
Adding to this, if we do move away from Apache HttpClient, we should probably do this across the board, even where we're not forced to be the deprecation of the osmdroid HttpClientFactory.  I imagine much of the complexity here will be around the `ApiClient` class due to the multipart entities it constructs.  Using the UrlConnection code natively may be a bit of a headache, so perhaps we should consider including a 3rd party library such as [Ion](https://github.com/koush/ion), [Loop4J](http://loopj.com/android-async-http/) or [Retrofit](https://square.github.io/retrofit/).  Any thoughts on that?

oliverlockwood:
[...]

PR #134, opened 2016-06-17
Move `ApiClient.getPOIs()` endpoints to Retrofit-based API client

oliverlockwood:
This is the first step in moving towards addressing https://github.com/cyclestreets/android/issues/133, by iteratively moving away from any dependency on the Apache HttpClient.

Key points worth highlighting:
-  By leveraging Retrofit/OKHttp/Gson, I believe the use of the underlying CycleStreets APIs becomes much clearer to the developer/maintainer; we also leave JSON parsing to a library, enabling us to remove a whole bunch of code.  (Note that we could equally use Jackson or similar for JSON conversion; Retrofit's documentation recommended Gson in the first instance so I went with that.)
-  For now, I am leaving the interface to the rest of the app as the static `ApiClient` class.  Once full migration is complete, we may want to revisit that, but I believe it makes sense to change things in small chunks.
-  Perhaps controversially given their previous absence 😉 , I've included unit tests of the new functionality. (Ideally, I would have liked to first create working unit tests of the old `ApiClient`, but it turned out that I'm not currently familiar enough with Android's intricacies to figure out how to get that working in a reasonable timescale.)

@jezhiggins @mvl22 This is ready for review when you have the time.  I'll hold off migrating the other `ApiClient` operations until the strategy has been fully agreed, to save us from rework.


mvl22:
Thanks, Oliver.

I don't know if this adds any value, or is worthwhile considering, but it might be worth seeing if there is a standard GeoJSON parsing library available, as all the API V2 output that is geographical (i.e. almost everything) will be in the standard GeoJSON layout - I notice all the manual definition of types like lat/lon etc, which presumably a library might do automatically.

In the case of Leaflet.js for the web, we just load the GeoJSON parser and it basically knows how to connect everything up automatically.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
```diff
@@ -0,0 +1,56 @@
...
+import retrofit2.Response;
+import retrofit2.Retrofit;
+import retrofit2.converter.gson.GsonConverterFactory;
+
+public class RetrofitApiClient {
+
+  private final V1Api v1Api;
+  private final V2Api v2Api;
+
+  public RetrofitApiClient(String v1Host, String v2Host, String apiKey) {
```
mvl22: I would set the order as `v2Host, v1Host, apiKey` or `v2Host, apiKey, v1Host` as V1 is basically deprecated and the least important of the three.

Also, the V1 API does run over HTTPS if you want, and there's no reason not to do that, so please update references from http://www.cyclestreets.net/ to https://www.cyclestreets.net/

oliverlockwood: Actually, I'd go one further and say that having a constructor with three `String` parameters is a bad smell.  Consequently, I've changed this to use a builder... see what you think.

Regarding updating http references to https, are you referring just to the `ApiClient` itself (https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/ApiClient.java#L65) or more widely?  If the latter, I think it warrants its own PR separate from this one.

oliverlockwood:
@mvl22 great call on using a library for GeoJSON parsing / POJO definitions;  I hadn't processed that the API response was standard GeoJSON, so thanks for the pointer.

PR updated.  With this in place we're able to remove some more boilerplate code, which is great.  Note that as a result of this I've switched to using Jackson instead of Gson with retrofit, as the GeoJSON library support for Jackson seemed to be better.

mvl22:
Thanks again, Oliver.

@jezhiggins Can you review?

Issue #121, opened 2016-03-27
Migrate geocoder to new /v2/geocoder API call


mvl22:
This is now available.

http://www.cyclestreets.net/api/v2/geocoder/

This brings the key benefit that you can return the full geometries, using geometries=1, so the map can be zoomed in, rather than merely be panned but stay at the same zoom level.

oliverlockwood:
Worth noting for the record here that the latest code in master now calls the V2 API.  We've not yet enhanced what we do with the response, in order to fully satisfy this issue, but it should be easier to address that now.

PR #135, opened 2016-06-25
Migrate `getPhotos()`, `getUserJourneys()`, and `geoCoder()` APIs to Retrofit

oliverlockwood:
Follow on from #134.  Contributes to the completion of #133.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,9 +61,7 @@ subprojects {
       exclude 'META-INF/NOTICE.txt'
     } // packagingOptions
 
-    // Required as per:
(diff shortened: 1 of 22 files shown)


oliverlockwood:
@mvl22 Having raised this, I've just noticed #121.  Is it worth changing this PR to address that issue in tandem?

jezhiggins:
Fully implementing 121 involves a bit more work that just migrating the geocoder call, but you could certainly do the endpoint if you wanted :)  (And more, if you like!)

mvl22:
@oliverlockwood Yes, please do move to V2. As Jez says, there is other work there (geometries=1) but I'd think about that stuff later.

Issue #138, opened 2016-06-29
Registration confirmation email sent by V2 API provides a broken link for validation


oliverlockwood:
I've tried to register a number of accounts - using my own API key against the REST endpoint, and through the app.  I receive the "Confirmation email has been sent" message, but no confirmation emails have been received.  As a result:
-  I can't verify any authenticated endpoints that I'm migrating to Retrofit.
-  I'm concerned there may be a wider issue affecting users.

(As a side note, is this the right place to raise issues with the CycleStreets servers themselves?)

Maybe I'm being impatient? but I have waited a couple of hours and I believe most users would expect an almost-instant email ✉️  I don't know what the expectation is, perhaps @mvl22 or @jezhiggins could enlighten me here.

mvl22:
I got a bounce from

`"youraddress@example.com"@localhost`

where your address is your actual address. I haven't seen this with the main website.

Can you check that you are sending the right data?

oliverlockwood:
@mvl22 I've just emailed you about this with my actual email address to make sure we're talking about the same thing.
The `@localhost` bit is weird, I've no idea where that would have come from...

mvl22:
Localhost would be added to the mailserver if it sees only a local part of the address.

I suspect you are sending a quoted string, i.e. `"youraddress@example.com"` rather than a string on its own, i.e. `youraddress@example.com`. Could you check?

`"youraddress@example.com"` would be interpreted I believe by
http://tools.ietf.org/html/rfc2822#section-3.4.1
as a local part because that seems to say that a @ in a quoted string would be treated as just any other character within the local part, not the @ delimiter.

On the other hand this could be a bug at the API end but as I say, I would expect to see that with other clients if so.

oliverlockwood:
Some progress on this one.
-  It seems like I might have messed things up in my early attempts to integrate with the create user API, and so subsequent attempts for the same email addresses didn't work.
-  I registered through the app for my `@sky.uk` email address, and that worked fine (using the V1 API, of course).  The email came through, I was able to validate the account and that's now being used to verify the "authenticate" API in integration testing.
-  Running out of email addresses, I registered through the integration testing of the V2 API for my wife's email address (which should be fairly obvious from your logs).  The email came through, however the validation link (https://api.cyclestreets.net/signin/register/<UID>/) says `404 https resource not found`.

[...]

PR #139, opened 2016-06-29
Migrate `register()`, `signin()`, and `sendFeedback()` APIs to Retrofit

oliverlockwood:
Similar to the other PRs I've raised recently.

A couple of things to note:
-  As part of migrating each of these, I've also upgraded to use the V2 API.
-  I haven't been able to integration-test the new implementation of the signin() API, due to #138.  If this can't be resolved swiftly, could you let me have a test email and password so I can verify this aspect?


oliverlockwood:
@mvl22 @jezhiggins another iteration of the Retrofit migration for review when you get the chance, if you don't mind.  Hope Wednesday's treating you both well.

mvl22:
More great work, Oliver - thanks.

I notice that this includes some tests (great) but that these seem to test comparison of the `test` key's value, as in
libraries/cyclestreets-core/src/test/resources/__files/registration-error.json
libraries/cyclestreets-core/src/test/resources/__files/registration-ok.json

Please be aware that, as documented, the string values that the API responds with may in future vary. For instance, when we shift the API to having internationalisation support that will definitely change.

So, to be safe, I would test for the presence of a non-empty string, or indeed just the presence of a key called `test` which should be enough to verify a test.

oliverlockwood:
Glad to help out 👍 

@jezhiggins - Thanks for merging.

@mvl22 Can you clarify your concern?  [RetrofitApiClientTest](https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-core/src/test/java/net/cyclestreets/api/client/RetrofitApiClientTest.java) (the only place where [registration-ok.json](https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-core/src/test/resources/__files/registration-ok.json) and [api-error.json](https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-core/src/test/resources/__files/api-error.json) are used) is a unit test so effectively the tests verify:
-  that whatever the server returns (in this case mocked by Wiremock) is propagated through the system
-  hardcoded strings in the app itself
I believe therefore that those tests are robust to changes in the string values returned by the server.

If I've misunderstood, please accept my apologies and point me more specifically at what needs changing 😀

mvl22:
Ah, thanks for the clarification!

PR #140, opened 2016-07-01
Migrate uploadPhoto() API to Retrofit

oliverlockwood:
Continuing the fun!

### DIFF
--- a/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
+++ b/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
@@ -9,8 +9,10 @@ import net.cyclestreets.api.POI;
 import net.cyclestreets.api.Photos;
 import net.cyclestreets.api.Registration;
 import net.cyclestreets.api.Signin;
+import net.cyclestreets.api.Upload;
 import net.cyclestreets.api.UserJourneys;
 import net.cyclestreets.api.client.dto.SendFeedbackResponseDto;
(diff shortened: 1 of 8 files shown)


libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
```diff
@@ -153,4 +159,29 @@ public UserJourneys getUserJourneys(final String username) throws IOException {
...
+                                   final String caption,
+                                   final String filename) throws IOException {
+    MultipartBody.Part filePart = null;
+    if (filename != null) {
+      File file = new File(filename);
+      RequestBody fileBody = RequestBody.create(MediaType.parse("multipart/form-data"), file);
+      filePart = MultipartBody.Part.createFormData("mediaupload", file.getName(), fileBody);
+    }
+    // Unfortunately we have to do all this faff, otherwise the JSON converter will insert quotes!
+    RequestBody usernamePart = RequestBody.create(MediaType.parse("text/plain"), username);
```
oliverlockwood: @mvl22 FYI, this was why we experienced some weirdness around email addresses - in my earlier efforts at migrating the user creation API (before [this commit](https://github.com/cyclestreets/android/commit/1ff3101c85de12e9318c99a2e67cb90736e39245#diff-1feaff77004e964c4335bda733e14addL33)) I was using multipart instead of form-urlencoded, and as a result the JSON converter was adding quotes around the strings.

oliverlockwood:
@jezhiggins when you get the chance, please review.  Many thanks.

PR #142, opened 2016-07-09
Finish ApiClient migration to Retrofit

oliverlockwood:
This PR comprises the following:
-  Migrate `getPOICategories()`, `getPhotomapCategories()` and `getBlogEntries()` APIs to Retrofit
-  Implement caching strategy for these three endpoints in a Retrofit/OkHttp manner.
-  Optionally read an API key from a git-ignored file in the integration test, so developers don't have to remember to remove their own API key before committing.
-  Tidy up.

Notes about caching:
-  As commented in the new `RewriteCacheControlInterceptor` class, the convention for determining whether an endpoint should be cached is by inspect the `Cache-Control` headers that come back from the server.
-  The V2 APIs return `Cache-Control:no-store, no-cache, must-revalidate` which indicates to the client that it should not perform any caching at all.  Interestingly, the `pois.types` and `photomap.categories` endpoints return a field in the *body* e.g. `"validuntil": 1469190075,` which gives a Unix timestamp 7 days in the future.
-  To achieve caching in this PR,we include the described interceptor as a network interceptor, so the `Cache-Control` header is rewritten before OkHttp assess it to determine whether or not it should cache.  Currently, the code in this PR ignores the `validuntil` value and just uses the existing hardcoded cache durations for the three endpoints that were already being cached.
-  In future we may want to consider updating how the server writes its Cache-Control headers, which would then obviate the need for some of this code.

@mvl22, @jezhiggins for review please.  Have a good weekend!


jezhiggins:
Good stuff.  Thanks.

oliverlockwood:
@jezhiggins my pleasure 👍 

Do you have any thoughts about using the "validuntil" field to determine our cache duration, or upgrading the server to generate more standard `Cache-Control` headers so clients don't have to figure them out for themselves?

mvl22:
Sounds as if we should set a proper Cache header whenever we send `validuntil` - would that resolve this?

Are there any other cases where you think the API should be sending a different `Cache-Control` header?

oliverlockwood:
@mvl22 Yes, that would resolve it and obviate the need for the interceptor in those cases.

I have raised #144 to cover the issue of the server response `Cache-Control` headers, so we don't have the discussion in a closed PR (which risks getting lost).

PR #143, opened 2016-07-12
Migrate remaining usages of Apache HttpClient

oliverlockwood:
As discussed previously, before we can migrate to the latest osmdroid library, and generally for getting our app up-to-date, we need to remove the remaining usages of `org.apache.http.*` classes.  These existed in the following places:

1.  `TileSource` - to initialise the `HttpClientFactory` to set the `http.useragent` to be `AppInfo.version()`
2.  `ImageDownloader` - for downloading CycleStreets images, using the `thumbnailUrl` field returned from the `photomap.locations` API.
3.  `CycleStreetsTileDownloader` and `MapsforgeOSMDroidTileProvider` - for getting map tiles (I haven't worked out what the differences between these are; my guess is one would use the downloadable map pack, but @mvl22 @jezhiggins I'd welcome some more background info on this).

My thought processes went like this:
-  Using Retrofit requires you to set a `baseUrl` when initialising the client object.
-  For `ImageDownloader` we could hardcode this to be `https://www.cyclestreets.net/` (as all the paths *currently* start with `https://www.cyclestreets.net/location/`) but that feels to me a bit hacky.
-  In any case, we couldn't do that for the OSM case because those produce dynamic baseUrls (see https://github.com/osmdroid/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/tileprovider/tilesource/OnlineTileSourceBase.java#L29-34).
-  To solve this, we therefore have two options.
  * Create Retrofit client objects on the fly.
  * Use OkHttp directly.

Since at the moment we're not doing anything with async on these requests, and because we have no prior knowledge of API structure, just a URL string that we're given, I see no advantage in the more heavyweight Retrofit wrapper for these cases and have therefore used OkHttp directly.

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
@@ -4,20 +4,14 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.preference.ListPreference;
 
-import net.cyclestreets.AppInfo;
 import net.cyclestreets.CycleStreetsPreferences;
 import net.cyclestreets.util.MapPack;
 import net.cyclestreets.util.MessageBox;
 import net.cyclestreets.util.Screen;
 import net.cyclestreets.view.R;
 
-import org.apache.http.client.HttpClient;
-import org.apache.http.impl.client.DefaultHttpClient;
-import org.apache.http.params.CoreProtocolPNames;
 import org.mapsforge.android.maps.MapsforgeOSMTileSource;
 import org.osmdroid.ResourceProxy;
-import org.osmdroid.http.HttpClientFactory;
-import org.osmdroid.http.IHttpClientFactory;
 import org.osmdroid.tileprovider.MapTileProviderBase;
 import org.osmdroid.tileprovider.tilesource.ITileSource;
 import org.osmdroid.tileprovider.tilesource.XYTileSource;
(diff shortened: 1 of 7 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
```diff
@@ -209,17 +203,6 @@ private static void addBuiltInSources(final Context context) {
 
   public static MapTileProviderBase mapTileProvider(final Context context) {
     addBuiltInSources(context);
-
-    final IHttpClientFactory httpFactory = new IHttpClientFactory() {
-      @Override
-      public HttpClient createHttpClient() {
-        final DefaultHttpClient client = new DefaultHttpClient();
-        client.getParams().setParameter(CoreProtocolPNames.USER_AGENT, AppInfo.version(context));
```
oliverlockwood: No need for this, we use an `Interceptor` now

oliverlockwood:
Note that the diff in at least one of the files is a little harder to read because there was a mixture of tab and two-space indentation, which I standardised to the latter as that seems to be what's used most frequently in the app codebase.

libraries/cyclestreets-view/src/main/java/org/osmdroid/tileprovider/modules/CycleStreetsTileDownloader.java
```diff
@@ -82,35 +77,53 @@ public Drawable getDrawable(InputStream aTileInputStream) throws LowMemoryExcept
...
+    client = new OkHttpClient.Builder().addInterceptor(builder.interceptor).build();
   }
 
-  public CycleStreetsTileDownloader(final ITileSource pTileSource,
-                                    final IFilesystemCache pFilesystemCache,
-                                    final INetworkAvailablityCheck pNetworkAvailablityCheck,
-                                    int pThreadPoolSize,
-                                    int pPendingQueueSize) {
-    super(pThreadPoolSize, pPendingQueueSize);
+  public static class Builder {
```
oliverlockwood: I switched to a builder because there was already a plethora of constructors.

oliverlockwood:
Ping @jezhiggins @mvl22 - any chance of this PR being reviewed, please?

jezhiggins:
Sorry chum, I thought I'd already done this one.

oliverlockwood:
:+1:

mvl22:
Report from a user trying the private builds:

`FYI the most recent version that doesn't crash on loading on a Samsung Galaxy s6 edge (android 6.0.1) is the following one: CycleStreets-vNext-2016-06-11-0043.apk`

(May not be related to the above work specifically, but perhaps another recent change.)

PR #145, opened 2016-07-21
Upgrade osmdroid

oliverlockwood:
Addresses #133.

There's not any tests for this functionality, and it feels like it's potentially more risky than the previous PRs I have raised.  @jezhiggins / @mvl22 / anyone else, can you help me out with this either by testing it yourself or, preferably, giving me advice / pointers on how I can test it myself?


libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
```diff
@@ -112,32 +112,23 @@ public static ITileSource createStandardTileSource(final String name, final Stri
...
+    return createXYTileSource(name, 256, ".png", baseUrls);
   } // createStandardTileSource
 
   public static ITileSource createDensityAwareTileSource(final Context context,
                                                          final String name,
                                                          final String... baseUrls) {
-    return createDensityAwareTileSource(context, name, ResourceProxy.string.unknown, baseUrls);
-  } // createDensityAwareTileSource
-
-  public static ITileSource createDensityAwareTileSource(final Context context,
```
mvl22: Does OSMDroid now have native high-density tile support, hence why this is being deleted?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/CycleMapView.java
```diff
@@ -1,6 +1,5 @@
 package net.cyclestreets.views;
 
-import net.cyclestreets.liveride.PebbleNotifier;
```
mvl22: Can you confirm that Pebble support has not been removed? I presume this is some kind of removal of a dependency that didn't actually need to be in this class.

libraries/cyclestreets-view/src/main/java/org/mapsforge/android/maps/MapsforgeOSMDroidTileProvider.java
```diff
@@ -19,6 +19,8 @@
 import okhttp3.Request;
 import okhttp3.Response;
 
+import static org.osmdroid.tileprovider.constants.OpenStreetMapTileProviderConstants.*;
```
mvl22: Can I just get a safety check that this doesn't end up importing the default OSM tile location? We should be using our own tileserver URL. At present this is just proxying the OSM one but in future will have its own renderer. Our own one also supports requests with the 2x resolution flag in the URL.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/CycleMapView.java
oliverlockwood: Confirmed - this class had a private instance field of type `PebbleNotifier` that was never used, so I cleaned it up.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
oliverlockwood: I believe so - see my [comment](https://github.com/cyclestreets/android/issues/133#issuecomment-232309900) in the related issue - but that's one of the things I want to validate through actual testing.

libraries/cyclestreets-view/src/main/java/org/mapsforge/android/maps/MapsforgeOSMDroidTileProvider.java
oliverlockwood: The only reason for this import static is that previously fields such as `TILE_DOWNLOAD_MAXIMUM_QUEUE_SIZE` were a part of the superclass.  In the latest version of `osmdroid` they have been abstracted out to a separate constants class, and the static wildcard import seemed the cleanest way to continue incorporating them.

I don't see any way that this static import could affect the tileserver URL that we're using.

oliverlockwood:
@mvl22 your comments responded to; I have been away on holiday hence the delay in my replies, sorry about that.

As per my top-level PR comment, I would like to test this live; to achieve that, some support from you or Jez would be appreciated.  What's the best way to run through that with one of you - email? Skype? instant messaging?  Cheers.

jezhiggins:
Hi Oliver, I'm just back from my holidays too.  Thanks for this.  Drop me a line a line at jez@jezuk.co.uk and we can sort out getting this out in the wild.

mvl22:
Can you copy me in at info@ - would like to try this out!

PR #147, opened 2016-09-09
Fix three silly bugs

oliverlockwood:
The following three bugs have been fixed.  They're all rather silly.
-  Hostnames passed to Retrofit need the scheme to be included.
-  Retrofit needs the context to determine its cache dir.
-  Don't attempt to use Retrofit before it's been initialised.

Now that these changes have been made, the app starts up once again when I tested it locally.

Apologies for breaking the builds; I'll endeavour to make sure it doesn't happen again!


oliverlockwood:
@mvl22 @jezhiggins a PR of repentance for your review when you have the chance.  Thanks!

PR #153, opened 2016-09-15
Add basic app startup tests using Espresso

oliverlockwood:
See https://google.github.io/android-testing-support-library/docs/espresso/index.html for details of Espresso.  It's the framework that was generally recommended by the helpful folks in the Android development community of practice where I work.

With these changes, **and either an emulator running or a real Android device connected**, you can run `./gradlew cyclestreets.vNext:connectedCheck`.  I have proved that this a valid test for app startup, since we know `master` is currently broken, and the test fails with just the changes in this PR, but passes once the bug fixes in #147 are merged in.

I propose getting this PR (and for that matter #147!) in as-is.  Subsequently we can look at automating the creation, starting and stopping of an emulator on CI and/or local builds, so that this functional test (and others which will follow) can then become part of the build process.

### DIFF
--- a/cyclestreets.vNext/build.gradle
+++ b/cyclestreets.vNext/build.gradle
@@ -3,10 +3,16 @@ evaluationDependsOn(':libraries:cyclestreets-fragments')
 dependencies {
   compile project(':libraries:cyclestreets-fragments')
   compile 'ch.acra:acra:4.9.0'
+
+  androidTestCompile 'com.android.support.test.espresso:espresso-core:2.2.2'
+  androidTestCompile 'com.android.support.test:runner:0.5'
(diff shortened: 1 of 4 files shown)


oliverlockwood:
@mvl22 @jezhiggins when you get a chance, a review of this and the linked bug-fix PR would be handy.  Cheers guys.

jezhiggins:
Good stuff.  If we can pair it up with something like 
https://wiki.jenkins-ci.org/display/JENKINS/Android+Emulator+Plugin
then we should be able to run smoke test as part of the build process.

oliverlockwood:
My thoughts exactly :smiley: I'll email you separately about adding plugins to Jenkins as there's a couple I'm interested in.

PR #162, opened 2016-10-12
Send route feedback asynchronously

### LINKED ISSUE #157: Feedback regression
The "Comment on this route" system currently fails when clicking "Upload!" with "There was a problem sending your comments: null"

Reported in different issue #71.

oliverlockwood:
Fixes #157.

### Before:
screen shot 2016-10-12 at 11 35 47

### After (with network connectivity):
screen shot 2016-10-12 at 11 29 02

### After (without network connectivity):
screen shot 2016-10-12 at 11 28 29

### DIFF
--- a/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/ApiClient.java
+++ b/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/ApiClient.java
@@ -94,10 +94,10 @@ public class ApiClient
     return retrofitApiClient.geoCoder(search, w, s, e, n);
   }
 
-  static Feedback.Result sendFeedback(final int itinerary,
-                                      final String comments,
-                                      final String name,
-                                      final String email) throws IOException {
+  static Result sendFeedback(final int itinerary,
+                             final String comments,
+                             final String name,
+                             final String email) throws IOException {
     return retrofitApiClient.sendFeedback(itinerary, comments, name, email);
   }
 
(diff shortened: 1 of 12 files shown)


mvl22:
If we could avoid 'geeky' terminology that mentions an API endpoint URL and IP that would be good. Just use a more friendly message like "The phone is offline so cannot connect" or "There was a problem reaching the CycleStreets system to submit this feedback. Please try again later."

In general if the phone is offline, we should state on the submission page that they need to go online first, as failure is obviously entirely predictable in this case.

oliverlockwood:
@mvl22 Thanks for reviewing. The error message printed here is a dump of the message contained in the exception that was thrown. Displaying this in its raw form is as per the original implementation (which presumably worked up until network on main thread was banned).

I entirely agree with you that we should give a friendlier non-geeky message, but to me it's a more general problem affecting many APIs; maybe we can consider it as a separate issue to be followed up on... what do you think?

mvl22:
> it's a more general problem affecting many APIs; maybe we can consider it as a separate issue to be followed up on

Yep - fine. Sorry, hadn't realised that was affecting other areas.

oliverlockwood:
(I used network connectivity for the example as it was a trivial way to demonstrate the error case :smiley:)

oliverlockwood:
No worries :+1: Do you want to raise an issue, linking this PR, so we can keep track?

PR #169, opened 2016-10-23
Remove all files in unused package net.cyclestreets.api.json

oliverlockwood:
This JSON parsing has been replaced with using the Jackson library.  I just realised there are no references to this package from outside it, therefore it's entirely redundant and can be removed.


jezhiggins:
Back in the old days (!) Android 2.3 used this library for JSON parsing but it wasn't available in 2.2.  I needed to do JSON parsing, while maintaining 2.2 compatibility so just pulled it in.  So, yes, now we're all switched over to Jackson this can go.

oliverlockwood:
Nice to know some history :smiley:
I'm feeling lucky as an Android latecomer that the libraries have some maturity at this point!

Issue #160, opened 2016-10-11
Move hardcoded strings into resource file


oliverlockwood:
Some strings are in a resource file, allowing them to be localised, as per e.g. https://github.com/cyclestreets/android/pull/158

Some strings are hardcoded, meaning they're always in English.  This isn't very nice to our non-English speaking friends, and we should fix it.

oliverlockwood:
@rodrigo-brito once we've addressed this, would you be able to provide the additional translations into Portuguese, please? Thanks!

rodrigo-brito:
No problem, I do.

rodrigo-brito:
Uma sugestão: Coloquem o label "hacktoberfest" em uma issue de tradução, esse mês está acontecendo um evento de contribuição no GitHub, em menos de 1 dia o app vai estar traduzido para mais de 10 línguas diferentes.

mvl22:
I'll get Transifex account set up.

oliverlockwood:
Good suggestion @rodrigo-brito.

I'm going to try to work through the extraction of hardcoded strings in the next few days so we can take advantage of this.  #164 was the first step in this direction.

A further thought on string identification arose, as I nearly ended up with a conflict of strings named `photo_upload` being defined in both `cyclestreets-view` and `cyclestreets.vNext`, which would have caused problems.  I wondered if it might be worth prefixing things in a module-specific and logical manner, along the lines of the convention proposed by http://jeroenmols.com/blog/2016/03/07/resourcenaming/, potentially with a prefix of `cs_core_`, `cs_view_`, etc?

@jezhiggins, @mvl22 - can I have your thoughts on this ASAP (ideally today, so I can start some of this on the train home) so we can get a strategy defined before I spend time doing the rest of the string extraction?

mvl22:
Side point: presumably `cyclestreets.vNext` is the actual library? I assume the naming came from a period of refactoring but that it might be better to fix this at some point.

Jez I think is best-placed to answer the main question. To a non-Android expert though, that link you sent looks sensible.

oliverlockwood:
`vNext` is the actual Android app itself (I don't know why it's called that - Jez?)

The other modules (`view`, `core` etc.) are libraries.

[...]

PR #170, opened 2016-10-23
Extract hardcoded text into strings.xml to enable internationalisation

oliverlockwood:
Continuing to work towards #160.


oliverlockwood:
@mvl22 @jezhiggins Once this PR has gone in I think we can kick-start the internationalisation process.  There may be one or two hidden strings still lying around but I guess those can be picked up through feedback.

mvl22:
Great work, @oliverlockwood! @jezhiggins, can you review?

Issue #171, opened 2016-10-25
Add translations


oliverlockwood:
CycleStreets is currently available only in English.  It would be great to support other languages.

The following strings files all need localisation:
-  https://github.com/cyclestreets/android/blob/master/cyclestreets.vNext/src/main/res/values/strings.xml
-  https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-core/src/main/res/values/strings.xml
-  https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-core/src/main/res/values/arrays.xml
-  https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-fragments/src/main/res/values/strings.xml
-  https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-track/src/main/res/values/strings.xml
-  https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-view/src/main/res/values/strings.xml

For each of the files described in the header comment of this PR, you will need to put a file named the same but placed in a corresponding `values-XX` directory that you create.  `XX` corresponds to the country code you're translating into.

A couple more details:
-  Only the _values_ of XML objects should be translated.  The `name` attributes should _not_ be translated as these are what the app uses to identify the text entry.
-  Any entry marked with the XML attribute `translatable=false` should, as the attribute suggests, not be translated - and in fact must not be included in the localised files at all, otherwise the builds will fail.

**Please check for errors locally, by running `./gradlew lint` or similar, to ensure that there are no errors before you submit your PR, as this will save everyone from wasted time.  If for whatever reason you can't do this, please ensure you check the status of automated build that runs when your PR is raised, and fix any errors promptly.**

auyer:
I can translate do Pt-BR

auyer:
I can help with Pt-BR

GiORg1:
Hello

I can translate them in Russian. Can you give me exact instructions how translation must be added?

LauraLaureus:
I can translate into Spanish, and also need accurate instructions. Only the strings between the tags or the name attribute as well?

TroyCode:
I can do this in zh-TW, and also need instruction

mvl22:
Thanks everyone for these offers of help!

In the first instance, please could you each create an account on Transifex, and then sign up at:
https://www.transifex.com/cyclestreets/cyclestreets-android/

and we will get the translations going there shortly.

LauraLaureus:
[...]

PR #172, opened 2016-10-25
Brazilian Portuguese Translation

auyer:
Brazilian Portuguese Translation related to Issue #171 

I assumed I should copy every file from the "values" folders into the new "values-pt" folders, but only the requested files were translated.

Tell me if the other files should be translated as well, like the  [arrays.xml](https://github.com/auyer/android/blob/master/libraries/cyclestreets-core/src/main/res/values-pt/arrays.xml) file, and I'll make another PR!

But I also believe other parameter files that were in the "value" folders would not need to be duplicated, depending on how the programming is made.

### DIFF
--- a/cyclestreets.vNext/src/main/res/values/strings.xml
+++ b/cyclestreets.vNext/src/main/res/values/strings.xml
@@ -11,7 +11,7 @@
   <string name="cyclestreets_blog_updates">Blog - updated</string>
 
   <string name="crash_notif_ticker_text">Unexpected error, please send a report …</string>
-  <string name="crash_notif_title">CycleStreets has crashed …</string>
+  <string name="crash_notif_title">CycleStets has crashed …</string>
   <string name="crash_notif_text">Please click here to help fix the issue.</string>
   <string name="crash_dialog_title">CycleStreets has crashed</string>
   <string name="crash_dialog_text">An unexpected error occurred forcing the
--- a/libraries/cyclestreets-core/src/main/res/values-pt/arrays.xml
+++ b/libraries/cyclestreets-core/src/main/res/values-pt/arrays.xml
@@ -5,122 +5,22 @@
     <item>Quilômetros</item>
     <item>Milhas</item>
   </string-array>
-  <string-array name="unit_values">
-    <item>km</item>
-    <item>miles</item>
-  </string-array>
   <string-array name="speed">
-    <item>Unhurried 16km/h (10mph)</item>
-    <item>Cruising 20km/h (12mph)</item>
+    <item>Sem pressa 16km/h (10mph)</item>
+    <item>Normal 20km/h (12mph)</item>
     <item>Rápido 24km/h (15mph)</item>
   </string-array>
-  <string-array name="speed_values">
-    <item>16</item>
-    <item>20</item>
-    <item>24</item>
-  </string-array>
+
   <string-array name="routetype">
     <item>Mais tranquila</item>
     <item>Balanceada</item>
     <item>Mais rápida</item>
     <item>Mais curta</item>
   </string-array>
-  <string-array name="routetype_values">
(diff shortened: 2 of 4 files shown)


oliverlockwood:
Hi @auyer.  Thank you for submitting the PR!  I will work with you to get this ready to go in.

As you can see from the red ❌ marked on this PR, the automated build that ran on your pull request failed, so there are some things that need correcting.  For future reference, it's always worth looking out for automated builds as you can use them to get feedback sooner than waiting for a human to write back 😃 

I will mark the things that need changing inline where I can or in follow-on comments otherwise.

libraries/cyclestreets-fragments/src/main/res/values-pt/dimens.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <dimen name="app_defaultsize_w">632.0dip</dimen>
+  <dimen name="app_defaultsize_h">598.0dip</dimen>
```
oliverlockwood: This file, as well as `colors.xml`, `styles.xml`, `dimens.xml` doesn't contain text that can be localised, so there's no need to duplicate them in the `values-pt` directory.  (If the `-pt` directory doesn't contain a resource, the default is used.)

Please therefore remove any untranslated files from your PR.

oliverlockwood:
Also @auyer I have realised that the `arrays.xml` file you mention *should* be translated - good spot!  I will correct that on the issue description.  Please would you include it as part of updating your PR?

Many thanks!

libraries/cyclestreets-core/src/main/res/values-pt/arrays.xml
```diff
@@ -0,0 +1,125 @@
...
+    <item>Balanced</item>
+    <item>Fastest</item>
+    <item>Shortest</item>
+  </string-array>
+  <string-array name="routetype_values">
+    <item>quietest</item>
+    <item>balanced</item>
+    <item>fastest</item>
+    <item>shortest</item>
+  </string-array>
```
auyer: Ok @oliverlockwood , another question : in this file, we have ```<string-array name="routetype">```  and ```<string-array name="routetype_values">``` with the same information, but one capitalized and the other not. It looks like only the first one should be translated, and the other will be used internally, is that right?

oliverlockwood: @auyer You're exactly right!  You can tell we're new to this translation thing 😄 

I've just pushed a commit which marks the `_values` arrays as `untranslatable`, so you can remove those from your submission.

auyer:
@oliverlockwood can you check if the translation is really causing the problem here? 
Cos I have checked every file, and all I believe none of the untranslatable strings still are.
I don't have a lot of practice with Travis or similar tolls (yet), but I solved the issues I could spot from 466c511  in the last commit, but the build still failed.

cyclestreets.vNext/src/main/res/values/strings.xml
```diff
@@ -11,7 +11,7 @@
   <string name="cyclestreets_blog_updates">Blog - updated</string>
 
   <string name="crash_notif_ticker_text">Unexpected error, please send a report …</string>
-  <string name="crash_notif_title">CycleStreets has crashed …</string>
+  <string name="crash_notif_title">CycleStets has crashed …</string>
```
oliverlockwood: @auyer please revert the change on this line.

auyer:
Ohh, I just saw that you merged it already. Do you still need me to fix or add anything ?

oliverlockwood:
@auyer no worries, I just did the remaining fix myself - seemed simpler that way :-D

Thanks again for providing the translation. Do try the app if you haven't already!

PR #174, opened 2016-10-27
Russian translation

GiORg1:
I've added russian translation of listed files. Please review and add my PR

### DIFF
--- a/libraries/cyclestreets-core/src/main/res/values-ru/arrays.xml
+++ b/libraries/cyclestreets-core/src/main/res/values-ru/arrays.xml
@@ -1,0 +1,129 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+  <string-array name="units">
+    <item>Километры</item>
+    <item>Мили</item>
+  </string-array>
+  <string-array name="unit_values" translatable="false">
+    <item>km</item>
+    <item>miles</item>
+  </string-array>
+
+  <string-array name="speed">
+    <item>Неторопливый 10миля/ч (16км / ч)</item>
+    <item>Запас хода 12миля/ч (20 км / ч)</item>
+    <item>Быстрый 15миля/ч (24км / ч)</item>
+  </string-array>
+  <string-array name="speed_values" translatable="false">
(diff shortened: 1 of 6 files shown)


libraries/cyclestreets-core/src/main/res/values-ru/arrays.xml
```diff
@@ -0,0 +1,129 @@
...
+    <item>km</item>
+    <item>miles</item>
+  </string-array>
+
+  <string-array name="speed">
+    <item>Неторопливый 10миля/ч (16км / ч)</item>
+    <item>Запас хода 12миля/ч (20 км / ч)</item>
+    <item>Быстрый 15миля/ч (24км / ч)</item>
+  </string-array>
+  <string-array name="speed_values" translatable="false">
```
oliverlockwood: Where the `translatable=false` attribute is present, the XML block should not be included *at all* in the localised files.  Sorry for not making this clear.  Please would you entirely remove all blocks from your PR where this attribute is present?

oliverlockwood:
Hi @GiORg1.  Thanks for raising this PR!  I have one item of feedback which needs addressing, as it's currently causing the build to fail.  Please would you fix this and then ping me to re-review?

Note that you can check the build status and view details of failures by clicking the "Details" link.  At the moment there's a red :X: giving you a clue that the PR needs some work 😃

libraries/cyclestreets-core/src/main/res/values-ru/arrays.xml
GiORg1: Okay I will do this. Please write about this on issue page for not repeating this error next time

oliverlockwood: Yup, I have already clarified my previous comment in this area to say "Any entry marked with the XML attribute translatable=false should, as the attribute suggests, not be translated - and in fact must not be included in the localised files at all, otherwise the builds will fail." 👍

oliverlockwood:
Merging this, I'll fix the build myself to move things forward.

PR #175, opened 2016-10-27
Spanish translation

LauraLaureus:
Spanish translation related to issue #171


oliverlockwood:
@LauraLaureus Thank you very much for taking the time to do the translations and submit this PR.  I have a number of items of feedback which need to be addressed before we can proceed.

1.  Please only translate the files specified in issue #171.  Other files should not be included in your PR.
2.  In your change you have removed the original language `arrays.xml` and therefore entirely broken the application.  Please put this back in.  To be clear: your PR should add 6 new files, and **change no existing files**.  Until the PR is in this state, we can't go any further.
3.  You have missed translation of the string `<string name="registration_error_prefix">Your account could not be registered.\n\n</string>` in `libraries/cyclestreets-core/src/main/res/values-es/strings.xml`.  As a result, the build is failing for having an incomplete translation.

oliverlockwood:
@LauraLaureus are you intending to fix this PR?  I can't merge it in its current state.

@AlexandreInsua do you have any feedback on individual translations?

LauraLaureus:
I'll fix this, but not in this moment. This night or tomorrow.

LauraLaureus:
@oliverlockwood check it again, I have removed the files I duplicated when I duplicate value folder in order to rename it value-es, translate the missed string value and downloaded again the array.xml

oliverlockwood:
@LauraLaureus thanks for the update.  There are still some problems to be fixed:
1.  There is still an incorrect move of `styles.xml` from `values` to `values-es` that should have stayed where it was - please revert this file move
2.  There are also a number of cases where you have included XML blocks with the attribute `translatable="false"`.  These blocks need to be removed entirely from the translated files, as this attribute indicates that the block should **not** be translated.

When you have fixed these problems, please run `gradlew lint` in the root of the repo and verify that the build passes.

As a piece of friendly advice, when you're submitting PRs to the open-source world, it's best practice to make sure that the project still compiles with your changes 😃   Some people can get quite grumpy with PRs that meet this standard!

LauraLaureus:
@oliverlockwood  Could you tell me what it's the problem this time? I'm not used to work with android. And less used to use this Online compiler :/

oliverlockwood:
Travis build failed on CycleStreets-track library. Without being at my computer to check, my guess is that it's your inclusion of a translation of `save_wide_blank` which I believe is untranslatable. Can you double check that and fix? I'll look later and see if there are other outstanding issues. Cheers!

LauraLaureus:
@oliverlockwood you were right.  Test was successful. Can you tell me how to add Travi to a repo? Maybe I will use it in the future.

oliverlockwood:
@lauralaureus thanks for fixing. For setting up Travis, https://docs.travis-ci.com/user/getting-started/ is a good starting point.

mvl22:
I see we have Spanish translations - thanks! But these don't seem to be present at:
https://www.transifex.com/cyclestreets/cyclestreets-android/

Is there some syncronisation issue here, or was that done before Transifex was set up?

oliverlockwood:
@mvl22 there's no automatic synchronisation between Transifex and Github.  (If nothing else that's made obvious because the `cyclestreets.vNext` module, which that Transifex page still references, has been renamed `cyclestreets.app`.)

We had some conversations a while back when I played around setting up [txgh](https://github.com/transifex/txgh) integration but in the end we decided not to bother on the basis that we're not changing strings or adding languages very often.

Given this PR was just about adding Spanish translations, if we have more to discuss on this subject, I suggest we raise a separate issue for tracking, otherwise the discussions will be hard to find in future.

PR #183, opened 2016-10-30
Italian translation added #171

mikylucky:
#171


oliverlockwood:
Hi @mikylucky, thanks for submitting this pull request! 

Can you please remove all the XML blocks which are marked with the attribute `translatable="false"`?

There should not be translations of these, and including them makes the build fail.

Thanks!

mikylucky:
Now should be ok!

mikylucky:
Now is ok ;)

oliverlockwood:
Nice work - thank you!

Issue #144, opened 2016-07-12
Server should provide meaningful `Cache-Control` header on cacheable endpoints


oliverlockwood:
As described in https://github.com/cyclestreets/android/pull/142, the V2 APIs don't return meaningful cache-control headers, preferring to send `Cache-Control:no-store, no-cache, must-revalidate`.

#### APIs which the Android app caches, which we *should* return meaningful cache-control for:

-  Some endpoints (`/v2/pois.types` and `/v2/photomap.categories`) return `"validuntil": 1469190075` in the JSON body - where the validuntil field is a Unix timestamp 7 days in the future.  It seems a no-brainer that these endpoints should return `Cache-Control: max-age=604800` (or possibly `Cache-Control: public, max-age=604800`  where 604800 is the number of seconds in a week.

#### Other APIs that I believe we *probably* should apply caching to:
-  The Android app (this project!) also caches the blog post API (`/blog/feed/`) for 24 hours.  This API doesn't currently return any `Cache-Control` header, but it could (though I appreciate that this may or may not be more debatable than the previous two!)
-  Image downloads, `https://www.cyclestreets.net/location/*` (e.g. https://www.cyclestreets.net/location/1330/cyclestreets1330-size640.jpg) - I don't see why we wouldn't apply caching to these endpoints, especially since images are probably the largest things a client will access from the server!

#### Other APIs that we *could* apply caching to
Though it depends on the details of how these endpoints are used, which I'm not fully up to speed with - so @mvl22, @jezhiggins and others may have a better view on these.
-  `/api/journey.xml?plan=<routestyle>&itinerary=1234` - the retrieve previous itinerary API - if the app is likely to reload an itinerary e.g. on sleep / restore or whatever.
-  `/v2/pois.locations`, `/v2/photomap.locations` - if any clients are likely to reload POIs *in the same geographic area*

mvl22:
Will get the proper cache headers in soon-as. Won't be before Monday though.

mvl22:
> APIs which the Android app caches, which we should return meaningful cache-control for

Both cache-control and expires are now sent in the response, whenever the `validuntil` token is present:

```
Cache-Control: public, max-age=604800
Expires: Sun, 23 Apr 2017 15:38:51 +0100
```

Could you try now and let me know if these seem correct?

oliverlockwood:
[...]

PR #185, opened 2017-04-18
Remove workaround for endpoints that now return valid Cache-Control headers

oliverlockwood:
As discussed in #144 we can now remove some of this workaround logic.


oliverlockwood:
@jezhiggins when you have the chance, do you want to review/merge this?

Issue #179, opened 2016-10-28
Publish beta builds using official Android methods


oliverlockwood:
We should do this.

We'll need to decide whether we want open or closed beta testing.
(see https://support.google.com/googleplay/android-developer/answer/3131213?hl=en-GB).

jezhiggins:
Sorry, this ticket has passed me by.  The cyclestreets app is set to open betas, because closed is such a massive pain in the arse to sign people up to, but it's not publicised.

mvl22:
Great; looking forward to seeing this in the beta store (or whatever it's called!).

oliverlockwood:
I believe this is now working.  Each new build should automatically go to the Play store as a beta.  Two iterations of version 3.7 were published there this morning and I've downloaded them on my phone successfully.

Closing this issue accordingly.

mvl22:
Great. Can you give a quick set of instructions as to how to sign up and get the beta?

jezhiggins:
See https://support.google.com/googleplay/answer/7003180?hl=en-GB

mvl22:
Thanks all for your efforts on this - really great that we now have this new official channel. I'll do a quick blog post on this soon.

oliverlockwood:
:+1:

mvl22:
I'll try and get this blog post done tomorrow. (If you want to help me expedite this, some screenshots and text would be great.)

oliverlockwood:
I'll aim to do the screenshots ASAP. 

I reckon it would be worth holding off publishing the blog post until the new release is live (hopefully early next week if there's no more major issues uncovered).

mvl22:
> I reckon it would be worth holding off publishing the blog post until the new release is live

I was thinking the opposite - it would be good to have more testers!

oliverlockwood:
I guess it depends on whether you consider your audience to be beta testers, or general users and potential users.

mvl22:
The blog post will be aimed at beta testers, yes.

mvl22:
Any chance of these "How to sign up for the beta" screenshots? I've got the others now - thanks.

oliverlockwood:
@mvl22 we've already given instructions for that above and by email to Simon (on your shared email address)...

mvl22:
Blog post now at:
https://www.cyclestreets.net/blog/2018/08/31/android-app-beta-programme/

Let me know of any clarifications or changes.

PR #187, opened 2017-05-24
Do CI builds on Travis

oliverlockwood:
This change should allow us to build the signed APK in CI on Travis (which we already use for PR builds).  The intention being that we would no longer need to use Jez's private build server for the CycleStreets Android app.

All the secret data is encrypted according to https://docs.travis-ci.com/user/encrypting-files/, so that only Travis can decrypt it.  (Note that we still need to keep a safe copy of our signing information elsewhere, in case we ever move away from using Travis for our builds.)

If people are generally happy with this approach, I would propose the next steps as:
-  merge this, and then I'll do some verification work on the signing of the APK etc.
-  decide how we publish the built artifacts
    -  we could copy them onto a server somewhere (as is done by the existing Jenkins builds), or...
    -  (my preference) we could just tackle #179, and publish them as open beta builds.

 @mvl22 and @jezhiggins, your thoughts please?


mvl22:
Open beta builds (do they end up here on GitHub?) are good. :)

jezhiggins:
Yea, like it and am happy to merge.  A rather cool next step (but please don't interpret this as any kind of instruction) would be to push the signed builds directly up to the Play Store as an alpha or beta build.

oliverlockwood:
Good stuff.  I just fixed the bash errors (thanks Travis for checking my PR quality!) and I think this is good to be merged.

@jezhiggins @mvl22 we're all in accord then: push signed builds directly up to the Play Store as alpha/beta.
However I'm stuck, because I need access to the Google Developer Console.  Although Jez shared with me the password a while back, Google's security is wanting more verification than that, as follows:

screen shot 2017-05-24 at 21 07 59

I'll be in touch less publicly to work through this with you.

oliverlockwood:
I'm aware that the CI build failed as it tried to decrypt one of the license files.  I'll try and look at this tomorrow.

oliverlockwood:
FYI, build https://travis-ci.org/cyclestreets/android/builds/236041036 is running now and looks like it's successfully done the license decryption phase.  I committed https://github.com/cyclestreets/android/commit/23e96efefcfe754bf694996441c4bd6804a4c98e straight to master to fix this.

Issue #152, opened 2016-09-15
Upgrade appcompat-v7 dependency in cyclestreets-view


oliverlockwood:
Currently we have the following dependency in `cyclestreets-view`:

``` gradle
  compile 'com.android.support:appcompat-v7:20.+'
```

We should upgrade this to e.g.

``` gradle
  compile 'com.android.support:appcompat-v7:23.4.0'
```

If you try to just make this change, compilation fails with the following stack of errors.  I haven't dug into this further yet.

> /cyclestreets-android/libraries/cyclestreets-view/build/intermediates/res/merged/release/values-ldltr-v21/values-ldltr-v21.xml
> Error:(3) Error retrieving parent for item: No resource found that matches the given name 'android:Widget.Material.Spinner.Underlined'.
> /cyclestreets-android/libraries/cyclestreets-view/build/intermediates/res/merged/release/values-v21/values-v21.xml
> Error:(13) Error retrieving parent for item: No resource found that matches the given name 'android:TextAppearance.Material.Inverse'.
> Error:(15) Error retrieving parent for item: No resource found that matches the given name 'android:TextAppearance.Material.Large.Inverse'.
> Error:(21) Error retrieving parent for item: No resource found that matches the given name 'android:TextAppearance.Material.Medium.Inverse'.
> Error:(28) Error retrieving parent for item: No resource found that matches the given name 'android:TextAppearance.Material.Small.Inverse'.
> Error:(201) Error retrieving parent for item: No resource found that matches the given name 'android:Widget.Material.ProgressBar'.
> Error:(203) Error retrieving parent for item: No resource found that matches the given name 'android:Widget.Material.ProgressBar.Horizontal'.
> Error:(211) Error retrieving parent for item: No resource found that matches the given name 'android:Theme.Material'.
> Error:(221) Error retrieving parent for item: No resource found that matches the given name 'android:Theme.Material.Light'.
> Error:(163) Error retrieving parent for item: No resource found that matches the given name 'android:Widget.Material.AutoCompleteTextView'.
> Error:(174) Error retrieving parent for item: No resource found that matches the given name 'android:Widget.Material.EditText'.
> Error:(6) Error retrieving parent for item: No resource found that matches the given name 'android:TextAppearance.Material.Button'.
> Error:(3) Error retrieving parent for item: No resource found that matches the given name 'android:TextAppearance.Material'.
[...]

PR #188, opened 2017-05-24
Update Gradle, android-tools, appcompat

oliverlockwood:
This was a complete pain, but I think we're there now.
See individual commit comments for details.

It should address #152 and enable moving forward in many ways.

Please do review the code for correctness, don't just take my word for it... I should have gone to bed a while ago!


mvl22:
Well done - that all looks like tedious stuff! I'll let Jez review the code.

jezhiggins:
Good stuff.

oliverlockwood:
Oh, and for the record: the reason I did this work at this point was because the Gradle plugin to publish to Play requires the newer version of either Gradle or the Android tools (I can't remember which) and once that was upgraded everything else got tangled.  Hence I did the upgrade work in isolation from adding in the publishing logic.

PR #191, opened 2017-05-30
Set up Play publishing plugin

oliverlockwood:


build.gradle
```diff
@@ -122,6 +128,17 @@ def signingKey(project) {
       }
     } // signingConfigs
 
+    playAccountConfigs {
+      defaultAccountConfig {
+        serviceAccountEmail = 'name@server.com'
+        pk12File = project.file('key.p12') // Will need to encrypt this in Travis, obvs
```
oliverlockwood: I need a service account / pk12 combination.  As I'm not the account owner on the developer console, I can't set this up.

mvl22: Are you able to clarify what this stuff is about, and what I need to do (or what privilege you would need to have added if that's possible)?

(Sorry for my lack of knowledge of this Android publishing side - it seems to change constantly!)

oliverlockwood: @mvl22 sure.  This is about providing API credentials for our CI builds (running in Travis) to publish beta artifacts to Google Play.  You could think of it like the API key that's used for hitting the CycleStreets APIs.

I tried to set this up by going to the Developers Console (http://play.google.com/apps/publish) and clicked Settings, then API Access.  I got this screen:
screen shot 2017-06-05 at 08 33 16

I suspect if you go to the same place as the `info@cyclestreets.net` account, you'd get the option to create/enable an API key.  You could then privately send that `.p12` file to me, and I can do the rest.

Does that give you enough to go on?

mvl22: Yes, I get this by the way:

screen shot 2017-06-06 at 00 13 45

mvl22: @oliverlockwood @jezhiggins I've e-mailed you both the details.

oliverlockwood:
@mvl22 we're making progress. That service account needs some permissions adding.  I've emailed you the details.

oliverlockwood:
I was working with @mvl22 on this today.  I'm going to merge this now and check what happens in Travis.

oliverlockwood:
Just checked: we've got a beta build up now.  Happy days.

Issue #186, opened 2017-05-16
Increase hard limit of 12 waypoints in 'Tap to Route'?


oliverlockwood:
The Tap to Route with waypoints feature is a great tool for planning a custom route rather than an A-to-B one.

A few things I've noticed:
1.  There's a hard maximum of 12 waypoints, as hardcoded [here](https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.java#L610).  The [page documenting the underlying API](https://www.cyclestreets.net/api/v2/journey.plan/) says "Information on the new V2 API is only available to Administrators at present."  The old [V1 API documentation](https://www.cyclestreets.net/api/v1/journey/) says "There can be at most 12 waypoints".  **I'd love to know if there's a reason for this, or whether there's any scope for this being relaxed, perhaps as part of a migration to the V2 API (which I could create a PR for if I had access to the documentation!)**
2.  ~~While you're on waypoints 1-11, you can press the "back" button on the UI to remove the last waypoint if you placed it in the wrong place.  However, once you've placed waypoint 12, the back button stops responding and you're stuck.  Which is really annoying if you've painstakingly created a route only to mis-click on the last waypoint.  This is just a bug and needs fixing.~~ This has been addressed by #226.

oliverlockwood:
@jezhiggins @mvl22 any thoughts on item (1) above?

I'll dig into investigating item (2) when I have the chance.

mvl22:
@si-the-pie should be able to clarify why there's a limit of 12 points.

There is still no date for the V2 API. However, if you switch away from the V1 XML to the equivalent V1 JSON output that will presumably be most of the work done, and should enable you at last to rip out the last of the XML parsing stuff.

oliverlockwood:
@si-the-pie - can you please share any clarification on why the 12 point limit?  I agree with @mvl22 that it seems restrictive - which is in fact how I came across this problem.

Clearly an app developer could knit together multiple 12-point routes but it'd be a bit of a pain, and if the restriction isn't necessary it would keep things simpler.

oliverlockwood:
@si-the-pie @mvl22 a year on, would you mind answering my query about the hard-limit of 12 waypoints, please?

mvl22:
[...]

PR #192, opened 2017-06-29
Migrate journey api to json

oliverlockwood:
This PR starts us off in the right direction for migrating the app's use of the CycleStreets Journey API from XML to JSON, as discussed [here](https://github.com/cyclestreets/android/issues/186#issuecomment-301905721)

To complete this work is going to be quite complex.  Currently:
-  The raw API XML is passed out of `cyclestreets-core` into the `cyclestreets-view` module.
-  `cyclestreets-view` parses the raw API XML into domain objects for use right now.  So far so good.
-  `cyclestreets-view` also **persists the raw XML** in an SQLite database.

In terms of design, what we have right now is hard coupling between the API and our own domain's persistence.

Here's what I propose as the way forward.
-  The Journey domain objects should be defined in a `domain` package in`cyclestreets-core` (in fact, most if not all of the other domain objects are already in here, although the package naming isn't yet explicit).
-  As per the other domain objects, there should be a way to create the domain objects (our internal representation) from the DTOs that represent the current form of the API.  That removes coupling.
-  Database persistence should be based on serialization and deserialization of the domain objects, not the DTOs used on the current version of the API.
-  A DB migration step ensures that existing people with saved routes can have those routes upgraded, rather than lost.
-  In a while (say 1 year?) after the first release including this functionality, I'd suggest that the migration support should be removed to clean up the codebase.

Note that I'm not going to implement any more until we have agreement, as it'll be quite a lot of work and I don't want to go off on the wrong track.  @jezhiggins, @mvl22, your thoughts please?


mvl22:
My general thought is that using GeoJSON as the native storage structure would be worthwhile, mainly as it's a standard that is unlikely to change, and it is what the API is about to move to. So I'm not sure about the idea of converting the received API response to an internal representation, as that representation might as well be GeoJSON anyway.

oliverlockwood:
@mvl22 I'm perfectly happy for GeoJSON to be used in the internal representation (and indeed for any input into the form of that internal representation).  Then we can map from the current API output to that representation, and it's only the mapping piece that changes in future.

If that had been done originally, then this change would be fairly simple.  We have the migration challenge now precisely *because* the internal representation was coupled to a particular form of the downstream API XML.  Hence my desire to decouple.

jezhiggins:
Oliver, is this still an active PR that I've just forgotten about, or have you done further work that would mean you want to raise a new one?

oliverlockwood:
Hi @jezhiggins, sorry for the delay, I was on a cycling holiday last week 🚵 

This is still an active PR with open questions as per the original comments and my response to @mvl22.  As a precursor to moving away from the XML journey API, and being able to significantly simplify and clean up our backend code, I believe we need to decouple the domain (how we store a journey) from the response that the API provides.  This PR is trying to stimulate discussion on that point but things went quiet from the end of June.

So if you can please give your thoughts I'd be grateful!

mvl22:
> So if you can please give your thoughts I'd be grateful!

Can you clarify what feedback you actually want?

This all seems sensible, so go ahead.

Only issue is what type of GeoJSON - the API V2 will have a particular layout that covers multiple journey types.

Why not simply store the current JSON layout API response and push it into the reader? That way it's the same as what the API produces and there is no conversion involved.

PR #202, opened 2018-05-13
Housekeeping, and start of material design work

jezhiggins:


.travis.yml
```diff
@@ -4,9 +4,9 @@ jdk:
 env:
   global:
     # Remember to keep this in sync with config.properties
-    - compileSdkVersion=23
-    - minSdkVersion=17
-    - targetSdkVersion=19
+    - compileSdkVersion=25
```
oliverlockwood: `toolsVersion` and `compileSdkVersion` do not currently match `config.properties`

Which is probably why the Travis build is failing...

build.gradle
```diff
@@ -63,21 +65,22 @@ subprojects {
     // Required due to strictness of the Android developer tools.  See lint.xml for full details.
     lintOptions {
       lintConfig rootProject.file('gradle/lint.xml')
+      abortOnError false
```
oliverlockwood: Do we really want to disable `abortOnError`?  I thought it would be a useful mechanism to keep us honest (and potentially catch some classes of bug).

If there's warnings/errors that need to be ignored, we can add specific exclusions (ideally each one documented) in https://github.com/cyclestreets/android/blob/master/gradle/lint.xml...

ciLicense.sh
```diff
@@ -3,10 +3,10 @@ set -ev
...
   echo "Running a non-PR build on master - set up signing info"
-  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.vNext/license.properties.enc -out cyclestreets.vNext/license.properties -d
-  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.vNext/signature.asc.enc -out cyclestreets.vNext/signature.asc -d
-  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.vNext/keystore.enc -out cyclestreets.vNext/keystore -d
-  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.vNext/play-api-key.p12.enc -out cyclestreets.vNext/play-api-key.p12 -d
+  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.app/license.properties.enc -out cyclestreets.vNext/license.properties -d
+  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.app/signature.asc.enc -out cyclestreets.vNext/signature.asc -d
+  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.app/keystore.enc -out cyclestreets.vNext/keystore -d
+  openssl aes-256-cbc -k $openssl_file_encryption_key -in cyclestreets.app/play-api-key.p12.enc -out cyclestreets.vNext/play-api-key.p12 -d
```
oliverlockwood: Shouldn't the decryption output files be moved from `.vNext` to `.app` as well?  Otherwise the Gradle tasks which make use of these won't be able to find them.

config.properties
```diff
@@ -8,10 +8,10 @@
...
 
-compileSdkVersion=23
-minSdkVersion=17
-targetSdkVersion=19
+compileSdkVersion=26
+minSdkVersion=23
+targetSdkVersion=23
 
-toolsVersion=25.0.3
+toolsVersion=27.0.3
```
oliverlockwood: As noted in `.travis.yml`, we need to get these files back in sync.

libraries/cyclestreets-view/build.gradle
```diff
@@ -6,7 +6,8 @@ dependencies {
   compile 'com.getpebble:pebblekit:3.1.0@aar'
 
   // In the main app project, we're already pulling in 23.x via the `acra` dependency.
-  compile 'com.android.support:appcompat-v7:23.4.0'
-  compile 'com.android.support:support-v4:23.4.0'
-  compile 'com.android.support:support-annotations:23.4.0'
+  compile 'com.android.support:appcompat-v7:27.0.2'
+  compile 'com.android.support:support-v4:27.0.2'
+  compile 'com.android.support:support-annotations:27.0.2'
```
oliverlockwood: It concerns me that we're pulling in `com.android.support:support-annotations:27.0.2` here, and `com.android.support:support-annotations:25.3.1` in the main app project.  Can we get these aligned (or perhaps remove the latter dependency entirely?)

Also, is the comment above these dependencies perhaps no longer relevant?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.java
```diff
@@ -0,0 +1,16 @@
...
+import android.content.pm.PackageManager;
+
+public class Permissions {
+  public static boolean verify(final Activity activity, final String permission) {
+    if (!hasPermssion(activity, permission))
+      activity.requestPermissions(new String[]{ permission }, 1);
+    return hasPermssion(activity, permission);
+  } // verify
+
+  private static boolean hasPermssion(final Activity activity, final String permission) {
```
oliverlockwood: Can we please correct the missing `i` typo in this method name?

libraries/cyclestreets-view/src/main/res/layout/locationbutton.xml
```diff
@@ -0,0 +1,50 @@
...
+          android:layout_height="wrap_content"
+          android:layout_gravity="bottom|center_horizontal"
+          app:layout_anchor="@id/padding"
+          app:layout_anchorGravity="bottom|center_horizontal"
+          android:layout_margin="16dp"
+          app:fabSize="mini"
+          app:borderWidth="12dp"
+          android:src="@android:drawable/ic_menu_search"
+          app:backgroundTint="@android:color/background_light"
+          />
```
oliverlockwood: Is keeping this commented-out code in deliberate?  If so, perhaps a comment explaining why it's kept would be useful.  Otherwise, I'd advocate removing dead code.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/PhotoUploadButtonOverlay.java
```diff
@@ -1,55 +0,0 @@
...
-import android.content.Context;
-import android.content.Intent;
-import android.content.res.Resources;
-import android.graphics.Canvas;
-import android.view.MotionEvent;
-
-import net.cyclestreets.PhotoUploadActivity;
-import net.cyclestreets.view.R;
-
-public class PhotoUploadButtonOverlay extends ButtonOnlyOverlay {
```
oliverlockwood: Is something in place to replace `PhotoUploadButtonOverlay` and `PhotoUploadActivity`?  Or are we removing this functionality?  It's not 100% clear to me.

jezhiggins: This has been dead for a while - replaced by PhotoUploadFragment

mvl22:
Thanks - great to see this modernisation work :)

Just to say we're quite close now to the design work being ready - the prototype that Patrick built in the summer has now been 95% through a designer. I'll circulate this for comment shortly.

.travis.yml
```diff
@@ -18,10 +18,10 @@ android:
     # The BuildTools version used by your project
     - tools
     - build-tools-${toolsVersion}
-    
+
     # if you need to run emulator(s) during your tests
     - sys-img-${ANDROID_ABI}-android-${targetSdkVersion}
```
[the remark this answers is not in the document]
oliverlockwood: This now addressed in https://github.com/jezhiggins/cyclestreets-android/pull/48 - if you merge that in, your PR will be updated.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.java
oliverlockwood: I addressed this also in https://github.com/jezhiggins/cyclestreets-android/pull/48.

libraries/cyclestreets-view/build.gradle
oliverlockwood: I addressed this also in https://github.com/jezhiggins/cyclestreets-android/pull/48.

build.gradle
jezhiggins: In the long term no.  However, due to a change in the way permissions are now handled, the linter wants us to put security exception handlers around the places where we use the location manager.

We should do this, but it needs a little more thinking about that simply slapping in a try catch block.  For example, if we can't GPS then liveride should be completely disabled.  

Rather than do that in a hurry, I decided making the linter a bit less strict was the expedient thing to do.

oliverlockwood: That rationale makes perfect sense.  However, rather than disabling `abortOnError` globally, we could instead choose to ignore the specific problems.  If you add the following snippet into `lint.xml`, it does the trick (I tested locally).

```xml
    <issue id="MissingPermission">
        <!-- TODO: https://github.com/cyclestreets/android/issues/204 -->
        <ignore path="**/RecordingService.java" />
        <ignore path="**/LiveRideService.java" />
    </issue>
```

Would you be happy to make this change instead?

oliverlockwood: In fact; I'll merge your PR and raise this as a mini follow-up, to save time 😃

oliverlockwood:
This looks good to me and we have a green Travis build, so I'm merging.

PR #206, opened 2018-05-22
Dependency upgrades

oliverlockwood:
A few upgrades to dependencies that the linter pointed out to me.  Nice and simple.


jezhiggins:
Good stuff.

PR #207, opened 2018-05-22
Disable emulator-based tests on Travis, due to instability

oliverlockwood:
Much to my disappointment, I've been unable to get the connected Android tests (using an emulator) to run reliably on Travis.

We need the builds to run on Travis as that's how we're publishing updated betas (which could of course then be release candidates.

Therefore this PR removes the relevant parts of `.travis.yml`.  I'll check back at some point in a few months and see if Travis have improved things in this regard, at which point I can refer back to this delta as a starting point.

In the meantime, whenever we have PRs reviewed and ready to merge, I'll take the action of running the emulator tests locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,7 @@ env:
     - minSdkVersion=23
     - targetSdkVersion=23
     - toolsVersion=27.0.3
-    # Emulator-specific
-    - emulatorApiVersion=23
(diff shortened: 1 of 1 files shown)


mvl22:
Thanks.

Obviously it would be worth passing on upstream any useful findings, so we can increase the chance of this working in future.

PR #209, opened 2018-05-25
Fix beta publishing

oliverlockwood:
Connects to #179.


oliverlockwood:
Merging this myself so I can see if we've got this working.

PR #210, opened 2018-05-25
Use number of commits on branch as the version code

oliverlockwood:
Connects to #179.

The Play store APIs only accept an APK submission if the versionCode (an integer) has not already been used. You can see an example failure here:

>```
>:cyclestreets.app:publishApkRelease FAILED
>FAILURE: Build failed with an exception.
>* What went wrong:
>Execution failed for task ':cyclestreets.app:publishApkRelease'.
>> com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
>  {
>    "code" : 403,
>    "errors" : [ {
>      "domain" : "androidpublisher",
>      "message" : "APK specifies a version code that has already been used.",
>      "reason" : "apkUpgradeVersionConflict"
>    } ],
>    "message" : "APK specifies a version code that has already been used."
>  }
>```

Therefore, versionCode has to be monotonically increasing.

The most logical way I could come up with for deriving a monotonically increasing number for our CI builds was to use the number of commits.  On the `master` branch (which is the only branch that will attempt to push to the Play store anyway), this number should never go down, assuming that we don't intend to rewrite history.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@
 sudo: false
 dist: trusty
 
+# Get the full commit history (required to derive our version code, see the root build.gradle)
+git:
+  depth: false
+
 language: android
 jdk:
   - oraclejdk8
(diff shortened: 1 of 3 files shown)


mvl22:
Great idea - thanks.

oliverlockwood:
:+1:

PR #216, opened 2018-06-27
[Work In Progress] Allow route sharing

oliverlockwood:
I want to test this locally before getting people to spend their time reviewing, but I want Travis to build it, so here's a PR 😄


oliverlockwood:
These changes are unnecessary, as there's already "Share this route" functionality embedded in the "Change route type" menu.

I'll raise an issue to make this more accessible, if appropriate with reference to the newer UI designs.

PR #219, opened 2018-06-29
Elevation / Itinerary improvements

### LINKED ISSUE #218: Improve Elevation graph
1.  Currently, axis labels are arbitrary based on the minimum and maximum values of elevation. (e.g. a route I'm looking at now has 7m, 164m, 322m, 480m).  It would surely be better to have (in this example) gridlines at 0, 125, 250, 375, 500m.
2.  I imagine we could easily also calculate total elevation gain / loss across the route, which is a useful stat for people planning routes.

oliverlockwood:
Closes #218.

In addition to the changes mentioned in that issue, I've also:
-  updated the itinerary overview to use Material style `textAppearance` as per the guide here: https://gist.github.com/webserveis/c8c7e001c617829caa61dae385113e00
-  changed the Imperial height display to be in feet, not yards (which I'm pretty sure is more usual).

### Screenshots before:
screen shot 2018-07-10 at 09 30 33
screen shot 2018-07-10 at 09 30 57

### Screenshots after:
screen shot 2018-07-10 at 09 01 40
screen shot 2018-07-10 at 09 25 44

### DIFF
--- a/cyclestreets.app/src/main/assets/whatsnew.html
+++ b/cyclestreets.app/src/main/assets/whatsnew.html
@@ -2,12 +2,14 @@
   <ul>
     <li>Migrating towards Material design</li>
     <li>Added translations: Deutsch, Español, Français, हिन्दी, Italiano, मराठी, Português, Русский, Türkçe</li>
+    <li>Show total elevation gain/loss in the Itinerary/Elevation views</li>
+    <li>Round max/min of elevation graphs to make them more readable</li>
   </ul>
 
   <h3>Release 3.5</h3>
   <ul>
-    <li>Fixed <strong>Comment on this Route</strong> so you can, indeed, send a comment.</li>
-    <li>Other small bug fixes.</li>
+    <li>Fixed <strong>Comment on this Route</strong> so you can, indeed, send a comment</li>
+    <li>Other small bug fixes</li>
   </ul>
 
   <h3>Release 3.4</h3>
--- a/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/ElevationProfileFragment.java
+++ b/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/ElevationProfileFragment.java
@@ -8,7 +8,6 @@ import android.view.ViewGroup;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
-import com.jjoe64.graphview.CustomLabelFormatter;
 import com.jjoe64.graphview.GraphView;
 import com.jjoe64.graphview.GraphViewSeries;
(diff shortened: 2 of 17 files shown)


oliverlockwood:
@jezhiggins I've now got this working (and tested reasonably thoroughly using the emulator).

Would you please be able to review it ASAP?  Ideally I'd like to get it merged today or tomorrow so the beta gets built and I can test it throughly on my long-weekend cycle touring that starts on Thursday 😃

mvl22:
Any chance of a screenshot or two of this new feature?

oliverlockwood:
@mvl22 absolutely!  In fact I'd draft-updated the main PR comment to include these, and a better description, but I had just failed to press "Update" 😄.  Now look above and you should see what you're looking for.

mvl22:
Great. Can I suggest:

1) Firstly, I would flattening out the graph a bit to be more rectangular: at the moment these look like extremely steep hills, rather than a rolling landscape. I think users should avoid having to make that mental leap of flattening out the image. Don't be afraid of whitespace being left on the screen!

2) Check that the algorithm for working out the vertical scale (e.g. 0-300m shown above) is sensible when the range is small. I I recommend setting a minimum of say 100m, For instance, if the range is only 0-10m, this would result in a very steep hill appearing for a rise of only a few metres - instead you still want to show e.g. 0-100m, but with everything showing in the 0-10 lowest segment with the 11-100 space above being empty.

3) I think the font of the introductory data table is too small. Once you've flattened out the graph, this will give you that space.

oliverlockwood:
@mvl22 Thanks for the prompt and detailed feedback.  In response:

### Point 2
The algorithm will always round the lower bound *down* to the nearest 100m below, and round the upper bound *up* to the nearest 100m above.  See https://github.com/cyclestreets/android/pull/219/files#diff-1dda86841edb2b1da38cc04e32459692R40.  Hopefully this makes it clear that your example case will be satisfied?

### Points 1 and 3
I did actually deliberately reduce the size of this text.  Here's a screenshot using the current code (before the changes in this PR) taken on my phone (Moto G 4G):


With the addition of an extra couple of lines to cover the altitude gain / loss, I believe the amount of screen remaining for the graph will be inappropriately small.  I also believe the level of wrapping of the route details text emphasises the fact that it's too large.

This is not an ancient phone (maybe 1 year old?).

On the other hand, I'll agree that the emulator screenshots do make the text look a little small.  **Perhaps on this point we can consider the change as-is and see what it looks like in the beta, then discuss before we make fully public the release containing these changes?**

mvl22:
Sure, let's just get this out there. It will be cleaned up anyway with the new design stuff.

jezhiggins:
Good stuff.

PR #224, opened 2018-07-17
Show phone and opening hours on POIs, if available

### LINKED ISSUE #97: [BikeHelp] Provide information about bike shops in addition to their names
If you're out and about and have a mechanical (for example), it would useful not only to see the locations of nearby bike shops (as the app currently does) but also to be able to see the address, phone number, website if these are available - for example, you might want to be able to ring them to see if they're open and if so whether they'd be able to replace a broken gear cable / chain / spoke / whatever.

It is my understanding that the V2 POIs API provides this information, so the dev work here would be as follows:
-  design how this would be displayed on the UI
-  pass through the information from the updated POIs API and display in the UI appropriately (when the API is updated, and if the info is available for the given bike shop).

Any comments or pointers on relevant areas of the code are welcomed, I haven't had the time to go digging yet :smiley:

oliverlockwood:
It's been a few years since I wanted to make this change, but better late than never 😄
Closes #97.

This PR:
-  changes how we call the server APIs to retrieve POIs, to also request `osmTags`
-  uses `osmTags.url` as a fallback if `website` is empty
-  derives phone and opening hours from `osmTags`
-  displays phone and opening hours on POI bubbles.

Some screenshots follow.

## Before
screen shot 2018-07-17 at 10 38 34
screen shot 2018-07-17 at 10 38 18

## After

screen shot 2018-07-17 at 10 33 31
screen shot 2018-07-17 at 10 32 37


oliverlockwood:
@jezhiggins @mvl22 your review of this proposed change would be welcomed.  Thanks.

mvl22:
Rather than dump a (potentially long) URL on the screen, why not just have the word "Website", which is a clickable link to the URL?

oliverlockwood:
@mvl22 that's a reasonable question.  Two thoughts:

1.  If we're doing that, what would you think to just having the POI name clickable as a link (if a website is available)?  That obviates the need for a separate link labelled "website".

2.  The way of displaying the website predates this PR, as you can see from the "before" pictures.  Given this, would you be OK with us raising the clickable-link improvement as a separate issue, to keep the complexity of each PR lower?

oliverlockwood:
#227 raised to track the potential improvement here.

PR #228, opened 2018-07-18
Add Material design icon library, use for navigate (liveride) and directions icons

oliverlockwood:
The library described at https://github.com/code-mc/material-icon-lib allows us to use Material design icons without having to keep PNG files embedded in our app.

My thoughts were that this would make maintenance of the app much easier going forward, and better for aligning with up-to-date styles.

I've started by adding in the library, and replacing the LiveRide and Directions icons with their Material equivalents.  Screenshots of this change follow; please ignore the highlight circles and focus on the two icons in the top right.

**If this approach is acceptable, I would propose gradually replacing most of the icons we use (excluding photomap markers!) with ones from this standard library.**

## Before:
screen shot 2018-07-18 at 15 22 16

## After:
screen shot 2018-07-18 at 15 01 57


oliverlockwood:
@jezhiggins your thoughts sir?

jezhiggins:
Yes, I like the look of this.

Issue #231, opened 2018-07-20
Navigation bar regression


oliverlockwood:
I spotted that the current beta has a regression in the top navigation bar.

### Before
<img width="272" alt="screen shot 2018-07-20 at 19 40 01" src="https://user-images.githubusercontent.com/6017680/43020180-219119ca-8c57-11e8-9f4e-b5f136a50a1c.png">
<img width="275" alt="screen shot 2018-07-20 at 19 39 53" src="https://user-images.githubusercontent.com/6017680/43020181-21ac5f78-8c57-11e8-8045-e645f22c8754.png">

### After
<img width="272" alt="screen shot 2018-07-20 at 19 41 58" src="https://user-images.githubusercontent.com/6017680/43020178-21580fcc-8c57-11e8-888e-31c53b2205e3.png">
<img width="270" alt="screen shot 2018-07-20 at 19 41 51" src="https://user-images.githubusercontent.com/6017680/43020179-2175ca44-8c57-11e8-8550-d28a547f0108.png">

### Digging into it
I used `git bisect` and, taking account for non-compiling intermediate commits (which predate when we enforced squash-merging of PRs), the following set of changes is to blame:
https://github.com/cyclestreets/android/compare/464d819...588f25e

From this, I suspect it's most likely to be a problem resulting from either:
-  updating the `compileSdkVersion` from `19` to `23`, or
-  upgrading the `com.android.support:appcompat-v7:20.+` dependency to `23.4.0`.

I currently know very little about how the top navigation drawer is drawn and animated.  I don't suppose @jezhiggins you might be able to help track down this regression or at least give me some pointers, please?

oliverlockwood:
I have been mulling over this... maybe it's time to bite the bullet and upgrade the navigation bar to material style, fully?

It may or may not be easiest to build the new navbar from scratch.  Just an idea.

jezhiggins:
I noticed it at the time I made the change, but decided it wasn't worth chasing at that point given imminent redesign.

I think this is the intended behaviour of appcompat - you can find the same arrow all over the place - and I have a suspicion that the behaviour we had before (which I liked!) was probably by accident. 

While the new behaviour is what we want in some places - Settings, Photo Upload - it isn't in others.  I think rebuilding the navbar is probably the way through this.

oliverlockwood:
[...]

Issue #214, opened 2018-06-18
Android 8.0 (API level 26) from August 1, 2018


mvl22:
"Google Play will require that new apps target at least Android 8.0 (API level 26) from August 1, 2018, and that app updates target Android 8.0 from November 1, 2018."

https://developer.android.com/distribute/best-practices/develop/target-sdk

How near are we on that?

oliverlockwood:
See https://github.com/cyclestreets/android/blob/master/config.properties#L13-L15.

At the time of writing this:
>```
>compileSdkVersion=26
>minSdkVersion=23
>targetSdkVersion=23
>```

Thanks for raising this issue.  There's clearly some work for us to do before November.

oliverlockwood:
When we update this, it might be worth raising the minSdkVersion to 24 as well, as this allows a whole bunch more useful language features (Java 8 e.g. default methods on interfaces, etc.)

mvl22:
> minSdkVersion to 24

I believe that takes us to Android 7.0 released on August 22, 2016. That seems quite recent to me, and I suspect will cut a significant amount of our current userbase out. I think we'd need a very good reason to do that beyond just coding purity.

oliverlockwood:
Fair point!

PR #239, opened 2018-07-23
Address deprecation warnings

oliverlockwood:
I wanted to clear up as many deprecation warnings as I could before attempting to bump the targetSdk as part of #214.

This PR:
-  Removes some unused files that I encountered
-  Addresses many deprecation warnings (see individual commit messages for details)

What I've specifically *not* addressed are:
-  deprecation warnings around use of GeoPoint in database access / writing, since that'll require some DB migration work (which we need to do anyway to stop using XML and start using JSON)
-  deprecation warnings around the MainNavDrawerActivity (as this needs separate work, see #231)
-  ProgressDialog deprecation warnings (because after some experimentation I realised that requires quite a bit of rework, which I consider to deserve its own PR).
-  NotificationBuilder related changes - again, I think these deserve their own PR.


oliverlockwood:
@jezhiggins last one of the day, I promise 😅

Please do have a read through when you have the chance.

oliverlockwood:
@jezhiggins 😁

oliverlockwood:
@jezhiggins feel free to have a look-see into your local history and find any other bits of work you've done but not submitted 😉

oliverlockwood:
Merging this now as @jezhiggins has approved.

PR #242, opened 2018-07-24
Set maximum and minimum zoom levels to more sensible values

### LINKED ISSUE #89: Map zoom level can go too high
The map zoom level on all the map options seems to enable infinite zooming. The result is either:

1) The style looks pixellated; or

2) (worse) the null tile, i.e. "More from CycleStreets soon" tile appears, as a result of HTTP requests pointlessly being made.

The zoom level should be limited on a per-style basis, using the numZoomLevels levels in getSupportedBaseLayers shown at: http://www.cyclestreets.net/cyclestreets.js .

I think even the vector style should have the same limitations on it also, as allowing infinite zoom beyond about level 20 gives the user the false impression that more data might appear. 20 is already very close anyway.

oliverlockwood:
Closes #89.

Screenshots show the maximum and minimum zoom levels now supported, which I think are not unreasonable.

Note that while I was dealing with this trivial change, I fixed up the follow-me button "on" highlighting to be clear again, which had been regressed.

screen shot 2018-07-24 at 14 24 44
screen shot 2018-07-24 at 14 24 03

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/CycleMapView.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/CycleMapView.java
@@ -57,6 +57,8 @@ public class CycleMapView extends FrameLayout
 
     mapView_.setBuiltInZoomControls(false);
     mapView_.setMultiTouchControls(true);
+    mapView_.setMaxZoomLevel(17);
+    mapView_.setMinZoomLevel(2);
 
     overlayBottomIndex_ = getOverlays().size();
 
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.java
@@ -54,7 +54,7 @@ public class LocationOverlay extends MyLocationNewOverlay {
     mapView_ = mapView;
 
     onColor_ = Theme.lowlightColor(mapView_.getContext());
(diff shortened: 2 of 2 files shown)


mvl22:
We use 19 rather than 17 - that does give useful extra detail. So please do increase this a bit.

oliverlockwood:
@mvl22 apologies if I've erred here; when I experimented locally I couldn't see extra detail, only blurring, at greater zoom levels.

Can you please give me an example location (ideally with screenshots) showing the difference between the reference you're using for zoom 17 and 19, so we can make a comparison?

mvl22:
Look how much clearer the walkways between these buildings are, and how the building names at what could be a destination are more plentiful.

screen shot 2018-07-24 at 22 33 43
screen shot 2018-07-24 at 22 33 54

oliverlockwood:
@mvl22 thanks for sharing, very helpful.

After some digging, I found a separate piece of code that limits the max zoom level we request from a tile source to 17: https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java#L120

I'd really like to understand from @jezhiggins if there's any reason why this is currently set to 17.  (Looking at git history doesn't give any clues.)  Performance?  Other issues?

In the absence of a reason, then we can bump both this limit and the corresponding limit I set in `CycleMapView` as part of this PR.

mvl22:
Definitely that 17 limit should indeed be removed. Tilesources should ideally have a zoom level set independently for each.

PR #244, opened 2018-07-24
Work around loss of location on save

### LINKED ISSUE #243: On pause/resume, map location gets set to somewhere in the arctic
When resuming, an additional unexpected pause/resume step takes place, and on this second pause the location is somewhere deep in the arctic circle.  It's always the same place: (85.051128, -180.0).

This bug has been present for a while; I've only just got round to raising the issue.  It might be since we updated `osmdroid`, but I'm not 100% sure.

I've done some extensive debugging, including:
-  putting in breakpoints is `osmdroid` - which, helpfully, caused the bug *not* to recur - so there's obviously some timing window
-   a local build of `osmdroid` with extra tracing enabled - the bug still occurred, but I got no further in understanding what caused it.

I have come up with a workaround, which I'll raise as a PR shortly.

oliverlockwood:
Closes #243.  For full description, see that issue.

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/CycleMapView.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/CycleMapView.java
@@ -107,8 +107,8 @@ public class CycleMapView extends FrameLayout
     if (paused_)
       return;
 
-    // Some process seems to somehow set the co-ordinates to (85.051128, -180.0) and then
-    // pause/resume. Despite extensive debugging, I've not been able to figure out the cause.
+    // Some process seems to somehow set the co-ordinates to (85.051128, -180.0) - the north pole!
+    // then pause/resume. Despite extensive debugging, I've not been able to figure out the cause.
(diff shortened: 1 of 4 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/CycleMapView.java
```diff
@@ -101,17 +106,28 @@ public Overlay overlayPushTop(final Overlay overlay) {
...
     if (paused_)
       return;
+
+    // Some process seems to somehow set the co-ordinates to (85.051128, -180.0) and then
+    // pause/resume. Despite extensive debugging, I've not been able to figure out the cause.
+    // As a workaround, ignore ridiculous (ant)arctic co-ordinates - I can't imagine anyone needing
+    // cycle navigation in such places!
+    if (abs(getMapCenter().getLatitude()) > 80) {
+      Log.d(TAG, "Working around weird location setting bug");
+      return;
```
oliverlockwood: This is the actual "fix".  Everything else is adding useful debugging lines, and minor refactors.

oliverlockwood:
@jezhiggins if you wouldn't mind?

PR #245, opened 2018-07-25
Increase max zoom level to 19; refer to a common value for this

oliverlockwood:
For discussion.

Follows on the discussions in #242; also relates to #89.

If there's no logical reason why the max zoom should be limited to 17 (which is currently an open question with @jezhiggins) then this PR should do the trick.


mvl22:
Be aware that different levels are possible for different tilesources. 19 is fine as a standard for all, but OpenCycleMap does for instance let you go further.

oliverlockwood:
Good to know @mvl22.

If @jezhiggins is happy with switching to 19, then we can have this PR as a start, and consider tweaking if necessary in future.

jezhiggins:
This is good for now, but I'll revisit when as part of updating the mapsforge library. It's max zoom depends on the specifics of the map data you're using, so we might need to propagate that value out of the tile source and up.

PR #248, opened 2018-07-31
Upgrade target version to API level 26

### LINKED ISSUE #214: Android 8.0 (API level 26) from August 1, 2018

oliverlockwood:
Closes #214.

See each individual commit for details of the relevant changes.

The main change that was actually *needed* was the creation of `NotificationChannel`s, as per https://developer.android.com/training/notify-user/channels.  Everything else is tidying up.

@jezhiggins, could you please review?


mvl22:
Thanks. Really great to get the codebase modernised with these sorts of changes.

PR #249, opened 2018-07-31
Migrate journey API from XML to JSON

oliverlockwood:
(Supersedes #192.)

Quite a big one, this.

At a high level:
1.  We now call the v1 JSON API, instead of the v1 XML API, for journey routing.
2.  We convert the received JSON into domain JSON.
3.  For upgrade of historical stored routes, we convert the old XML into the same domain JSON format.  See `JourneyStringTransformerTest` for proof of this, with the resource files used as examples.
4.  There are now tests of the database upgrade process.
5.  The last deprecated usages of the `E6` forms of `GeoPoint` have been removed.

This work has enabled significant clean up of some convoluted XML parsing.

At some point in the future I imagine we will want to upgrade to the v2 JSON API when it's ready.  By decoupling our domain view from the API representation, this should be a much easier process.


oliverlockwood:
@mvl22 while I'm looking at this... I don't suppose there's any news on the V2 Journey API, is there?

oliverlockwood:
@mvl22 FYI.

@jezhiggins if you would be so kind to review, I'd be grateful!

PR #250, opened 2018-08-06
Update Android plugin to latest patch release; update Gradle

oliverlockwood:
Fairly trivial.

@jezhiggins your eyes, please?  I've checked through running the emulator, and there are no obvious regressions.


jezhiggins:
I've been running it on an unmerged branch and I've had no problems either.

PR #251, opened 2018-08-07
Update the navigation drawer

### LINKED ISSUE #231: Navigation bar regression

### LINKED ISSUE #233: LiveRide button not always shown when it should be
To reproduce:
1.  Plan a route; LiveRide button appears.
2.  Go to another section of the app e.g. the Photomap.
3.  Return to the journey planner.  The planned route is still there; the LiveRide button is not.

If you click the top right drop down menu, the LiveRide button re-appears.  So it's not super-critical but it is a bit irritating.

As far as I can tell this bug has been present for a little while, but I haven't pinned down an exact version at which it was regressed (or whether it ever worked properly at all!)

oliverlockwood:
Closes #231 (by significantly updating the navigation drawer to a modern style).
Closes #233 (I'm not sure exactly where that bug was, but it's no longer seen with this update).

Key changes:
1.  Navigation view is now Material style, overlapping the action bar, with a nice header image.  See screenshots below.
2.  Code is a lot simpler (look at lines added vs lines deleted).
3.  Lots of unused files deleted.
4.  The image you see on first load is higher-resolution than it was before.

### Menu
screen shot 2018-08-21 at 22 10 08

### Splash screen
screen shot 2018-08-21 at 22 10 00

### Menu
screen shot 2018-08-21 at 22 44 26

### Splash screen
screen shot 2018-08-21 at 22 01 11


oliverlockwood:
@jezhiggins @mvl22 This is now ready for review.

Quite a bit of work, but in my view a huge improvement that we should roll into the release so as to avoid the regression that we would otherwise have.

oliverlockwood:
@jezhiggins if you have a minute spare, it might be worth getting this change in before @si-the-pie does his pre-release testing 😃

Issue #46, opened 2012-02-25
List of supported countries


mvl22:
An API call will be available shortly giving a list of the countries that the CycleStreets API will route in.

It would be good to list these in the About/Settings area.

oliverlockwood:
@mvl22 is this API call available yet?

mvl22:
We do now have a prototype API call, but we need to implement a few things following a policy discussion last night. Will let you know when this is ready.

oliverlockwood:
Also, after reading https://support.google.com/googleplay/android-developer/answer/7550024?hl=en I was able to find out that we're currently available (according to the Android Play Store) in UK, Ireland, Netherlands & Czech Republic.

So when that API call is available and we want to include its information, we also need to make sure we keep those settings in the Developer Console in synch.

oliverlockwood:
@mvl22 ping

oliverlockwood:
I've now switched this to be labelled as "Server issue".

mvl22:
NB Is there an Android methodology in terms of country IDs that we can adopt to work this automatically?

oliverlockwood:
Email sent.

I think the most important thing is to return the response with a list of standard 2-letter country codes (perhaps in a tuple alongside their user-friendly names).

I can't find any evidence of being able to programmatically *set* the availability, but we might be able to programmatically *check* the Play Store availability in our builds.

The closest I've got in 5 minutes (without having to pay money to anyone) is to hit e.g. https://play.google.com/store/apps/details?id=net.cyclestreets&gl=gb from an incognito browser window; if you then change `gb` to something else e.g. `de`, `fr`, the Install button disappears.  So it's doable at the very least in a hacky way 😄

mvl22:
We have an API call which gives a GeoJSON of the exact routeable areas.

The ideal is that when the user clicks on a map location in the journey planner in an area that is outside these GeoJSON zones, there should be a modal dialog stating that “Routing is not available in this location.”

oliverlockwood:
@mvl22 can you please share details of that "routeable areas" API call - either here or in an email to me?
I can certainly use that info to create the modal dialog you describe (which I would have found useful when in Austria last September!)

In terms of a list of countries, then, I can see a number of options:
1.  Expose another API which gives a list of countries (as suggested by the original description of this issue)
2.  Dynamically generate a list of countries from the "routeable areas" API
3.  Manually work out a list of countries from looking at the API output
4.  You post a list of countries here, and we manually update it every now and again.

[...]

PR #252, opened 2018-08-08
Tidy up strings, etc.

oliverlockwood:
This PR:
1.  Removes the Hindu and Marathi translations.
   We only have these as a result of a contribution during Hacktoberfest.  The more translations we have, the more effort it is to maintain the app as we continue to refactor and improve.  From https://github.com/cyclestreets/android/issues/46, I strongly suspect that CycleStreets supported locales are not any time soon going to include countries where these are major languages.
2.  Reorganises some string resources.  This reduces some duplication between modules, removes translatability of the app name itself (which was never translated!) and will help with subsequent work on the menus, where we need to access some strings in a library instead of the top-level app.
3.  Updates the What's New page according to many recent PRs that have gone in.
4.  Fixes a minor bug where the route highlight buttons stayed around when the route was cleared.  Whoops.

@jezhiggins... would you mind doing the honours, please?


mvl22:
> strongly suspect that CycleStreets supported locales are not any time soon going to include countries where these are major languages

We do actually have coverage of some areas of asia...

That translation is OK to remove if you want though.

PR #253, opened 2018-08-09
Compile against new API level 28

oliverlockwood:
Compile using the newly released API level 28.

The major change that affected us is the deprecation of the `android.app.Fragment` and related classes in favour of their support library equivalents.  See e.g. https://developer.android.com/reference/android/app/Fragment.

This meant we had to make some tweaks to our `SettingsFragment` to get it to work; while I was there, I improved the transitions between settings pages to a more modern feel.


oliverlockwood:
@mvl22 and @jezhiggins - look how modern we're being; API 28 was only released 3 days ago 😃 

Jez, would you mind reviewing this please?

PR #254, opened 2018-08-09
Start using kotlin

oliverlockwood:
A simple migration of a couple of classes to Kotlin, as @jezhiggins and I discussed in detail some time ago.

What do you think, Jez?


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.5-bin.zip
```
oliverlockwood: Unrelated change, but Android studio was grumbling that Gradle was too advanced for it.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/JourneyStringTransformer.kt
```diff
@@ -0,0 +1,68 @@
...
+        val chainr = Chainr.fromSpec(chainrSpecJson)
+
+        val transformedOutput = chainr.transform(inputJsonObject)
+        return JsonUtils.toJsonString(transformedOutput)
+    }
+
+    private const val V1API_XML_JOLT_SPEC = """[{
+  "operation": "shift",
+  "spec": {
+    "markers": {
```
oliverlockwood: Multiline strings are one nicety we're taking advantage of by using Kotlin here.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
```diff
@@ -141,20 +141,20 @@ public String getJourneyJson(final String plan,
                                final String arriving,
                                final int speed) throws IOException {
     Response<String> response = v1Api.getJourneyJson(plan, itineraryPoints, leaving, arriving, speed).execute();
-    return JourneyStringTransformer.fromV1ApiJson(response.body());
+    return JourneyStringTransformer.INSTANCE.fromV1ApiJson(response.body());
```
jezhiggins: I think we can avoid this .INSTANCE. business by putting the fromV1APIJson method inside a companion object.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/JourneyStringTransformer.kt
jezhiggins: Which is a fairly large win here :)

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
oliverlockwood: Cool, I'll look into that.

oliverlockwood:
In my day job I've been spending a lot more time with Scala and Groovy lately.  It is a bit painful going back to writing getters and setters 😄 so let's see what more we get from Kotlin.  I'll address your comment and go a bit further with this in the next few days.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
```diff
@@ -132,7 +132,7 @@ public String getJourneyXml(final String plan,
                               final String arriving,
                               final int speed) throws IOException {
     Response<String> response = v1Api.getJourneyXml(plan, itineraryPoints, leaving, arriving, speed).execute();
-    return JourneyStringTransformer.fromV1ApiXml(response.body());
+    return JourneyStringTransformerKt.fromV1ApiXml(response.body());
```
oliverlockwood: @jezhiggins I went one better than the companion object idea, and just made [package-level functions](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#package-level-functions) in this case.

jezhiggins: Ah, good one - because all the methods on JourneyStringTransformer are static, we can do away with the class definition entirely. Nice.

Issue #258, opened 2018-08-11
"Take photo" crashes


oliverlockwood:
There's a toast message that says "Could not open uri" and then this stacktrace:

```
08-11 07:40:39.379 19482 19482 E ACRA    : ACRA caught a RuntimeException for net.cyclestreets
08-11 07:40:39.379 19482 19482 E ACRA    : java.lang.RuntimeException: Unable to resume activity {net.cyclestreets/net.cyclestreets.CycleStreets}: java.lang.RuntimeException: You must pass an IMyLocationProvider to setMyLocationProvider()
08-11 07:40:39.379 19482 19482 E ACRA    :     	at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3124)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3155)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1401)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at android.os.Handler.dispatchMessage(Handler.java:102)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at android.os.Looper.loop(Looper.java:148)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at android.app.ActivityThread.main(ActivityThread.java:5459)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at java.lang.reflect.Method.invoke(Native Method)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:728)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:618)
08-11 07:40:39.379 19482 19482 E ACRA    : Caused by: java.lang.RuntimeException: You must pass an IMyLocationProvider to setMyLocationProvider()
08-11 07:40:39.379 19482 19482 E ACRA    :     	at org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay.setMyLocationProvider(MyLocationNewOverlay.java:197)
08-11 07:40:39.379 19482 19482 E ACRA    :     	at org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay.enableMyLocation(MyLocationNewOverlay.java:535)
[...]

PR #260, opened 2018-08-13
Refactoring the "Add Photo" Fragment

### LINKED ISSUE #259: Add photo: After upload, a screen rotation causes the location URL to be lost
This is because `uploadedUrl` isn't currently persisted on pause / restored on resume.

Fix coming shortly as part of a bigger refactor to this area.

oliverlockwood:
This change should, I believe, have only one behavioural change: such that it closes #259.

The main work is refactoring the `UploadFragment`:
-  a separate package to contain all the code
-  Kotlin instead of Java, general tidying up accordingly
-  some static classes and helper functions moved out into separate files, to improve readability and separation of concerns
-  some renames etc. for readability & clarity

I'd like this to go in as a baseline from which I can then work on resolving #258 (since the error message and fallback to image select pathway, if not the crash, is 100% reproducible in the emulator as well as on my phone).

@jezhiggins would you please review when you have the chance?  Many thanks.


jezhiggins:
Will do. Sorry for not picking this up - very busy week this week.

oliverlockwood:
No worries!

PR #265, opened 2018-08-15
Show `Contacts` button in Find Place views, dynamically request permission when clicked

### LINKED ISSUE #195: Request permissions for contacts on-demand
The app has route-to-contacts feature, so people can for instance get the route to a friend's house.

Currently I believe the app requests general contacts permissions up-front at the time of installation from the app store. Users wonder why this is needed, potentially reducing trust. We've certainly had support requests as to whether we upload contacts to servers (which we don't, and have no support or desire to do).

We should instead move to the on-demand model, where permission is requested at the point that the permission is needed.

As stated in the privacy policy:

`CycleStreets apps that request access to the user's address book have this functionality purely so that the user can do 'route to address'. Apps do not, and are unable to, upload any such data to the CycleStreets server. This can be independently verified by viewing the source code of our apps.`

oliverlockwood:
Closes #195.

For review by @jezhiggins.


jezhiggins:
👍

mvl22:
Thanks so much for getting this sorted. The number of feedbacks we’ve had over the years asking for why the app needs contact permissions, and how many times I’ve had to respond that we don’t scrape addresses...

mvl22:
Please ensure this feature is added to the release notes.

Issue #204, opened 2018-05-22
If location permission not granted, completely disable affected features (e.g. LiveRide) 


oliverlockwood:
This came out of discussions on #202.

jf1:
I frequently have location disabled so hope there'll be a straightforward way to access those features if location is enabled once Cyclestreets is running, ie please don't disable then not re-check until restart, or add a manual way to trigger a re-check.

oliverlockwood:
@jf1 this isn't about location enabled/disabled, it's about the *permission* for an app to use location information.  I'd be surprised if anyone frequently revokes/allows such permission 😃

oliverlockwood:
Note that after #265 has gone in, the RouteMapFragment will remain be the only place that requests permissions at fragment startup, rather than when we attempt to use a particular feature.  I'll look into sorting this (alongside the specific LiveRide enabling/disabling issue described here) soon.

oliverlockwood:
#269 deals with the location permission.

That leaves us with the write data permission, which the `RouteMapFragment` still demands at startup time (for saving map tiles).

I'll look into what we can do here.

PR #269, opened 2018-08-17
Permissions: on-demand for location; provide justifications

oliverlockwood:
Connects to #204.

This PR:
-  means that location permissions are requested on-demand, rather than upfront
-  if a permission is denied by the user once, then subsequent times the user is first presented with the reasons *why* we would like the permission, before making their choice (and being able to suppress future requests).

Some screenshots to show the experience:

### At the first time of clicking LiveRide:
screen shot 2018-08-17 at 13 57 53

### At the second time of clicking LiveRide, if you clicked Deny the first time
screen shot 2018-08-17 at 13 58 12
screen shot 2018-08-17 at 13 58 19

@mvl22 FYI
@jezhiggins for review, please?


mvl22:
Is it really recommended that a technical Constant ACCESS_FINE_LOCATION be shown to the user? We should avoid this kind of geekery unless it is required.

mvl22:
Can the dialog in the first screenshot have the reasoning shown in the second screenshot? Not sure why these really need to be different - in both cases we should be providing information to reassure the user.

oliverlockwood:
Good call on hiding the geekery. I'll change this accordingly. 

The answer to your second question is no. The actual "request permissions" dialog is presented by the OS, and for security reasons there's no way of affecting how it looks.

The Android guidelines are:
-  it should be as intuitive as possible from the context as to why a permission is being requested (e.g. in the case of location, when the user clicks on the crosshairs or the LiveRide button)
-  if a user clicks deny the first time, it's recommended that the next time you ask, you first present your reasons before invoking the Android dialog (as from the second attempt onwards, the user can check a "Don't ask again" box

oliverlockwood:
@mvl22 with the latest push, the explanation dialog now looks like this.  What do you think of the new form?
screen shot 2018-08-17 at 21 11 49

Note that I've refrained from trying to give a verbose description of the permission's name, as the Android OS dialog does that already.

PR #271, opened 2018-08-18
More pre release work

### LINKED ISSUE #267: ACRA reports don't reach developers
We have ACRA crash reporting set up.

As a developer, I'm not seeing these crash reports, even when I've put comments in there saying "please forward to <my email address>".

We should look into the configuration of how these reports are submitted, and consider updating it.  In any case I believe the ACRA dependency we have is quite old, and we should update this as part of that work.

oliverlockwood:
-  Fix an exception, spotted using the Android Developer Console, that was plaguing one recent user
-  Remove ACRA as the above proves it's no longer relevant (and therefore, this PR closes #267)
-  Bump the versions of Jackson (trivial) and GraphView (some code changes required)
-  Add credits for all the other libraries we now use at run-time


oliverlockwood:
@jezhiggins would you mind, please?

oliverlockwood:
@mvl22 links now changed to HTTPS where possible.

PR #273, opened 2018-08-20
On demand write external permissions

### LINKED ISSUE #204: If location permission not granted, completely disable affected features (e.g. LiveRide)

oliverlockwood:
Closes #204, and hopefully finalises our permissioning improvements.

Key aspects:
1.  If write-external not yet granted, use app-internal storage to save map tiles (so the app is still usable).
2.  When write-external granted, shift to the normal map tile cache location.
3.  Make sure various fragments don't try to auto-enable location, unless permission has already been granted.
4.  Update the icon for "add location", while we're at it, to make it more Materialy.

### Previous "add location" icon
screen shot 2018-08-20 at 11 15 01

### New "add location" icon
screen shot 2018-08-20 at 11 09 05


libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleMapFragment.kt
```diff
@@ -0,0 +1,150 @@
...
+open class CycleMapFragment : Fragment(), Undoable {
+    private var map: CycleMapView? = null
+    private var forceMenuRebuild: Boolean = false
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, saved: Bundle?): View? {
+        super.onCreate(saved)
+
+        forceMenuRebuild = true
+
+        checkPermissionNoMoreThanOnceEveryFiveMinutes()
```
oliverlockwood: Because of the fragment lifecycle, if we don't do this then at the time of starting up the fragment we enter an infinite-request loop.  I think we should still ask for the permission each *meaningful* time we navigate to a map fragment (of whatever type) - hence this mechanism, rather than e.g. only asking at app startup.

Issue #277, opened 2018-08-24
Photomap - large icons


oliverlockwood:
Raised by @si-the-pie, by email:

Photomap some of the icons are the large version as shown in the attached screenshot.

![screenshot_20180823-162933_cyclestreets](https://user-images.githubusercontent.com/6017680/44586067-54910500-a7a6-11e8-9576-c7c32b369091.jpg)

oliverlockwood:
I have a fix for this that I'll be raising shortly.

The problem is in the code path where we fall back to a default image.  In this case, it seems we're missing the photomarker assets for the `bikeshare` category, which I'm guessing has been added as a new category since the app was first developed.

With the fix, our defaulting behaviour will be better, but I presume it would be nicer if we could include the new icons (which I assume have been created) for the `bikeshare` category.

To go with the others in https://github.com/cyclestreets/android/tree/master/libraries/cyclestreets-core/src/main/assets/photomarkers, we'll need 6 icons: good, bad, and neutral, in normal (38x42) and 2x (76x84) resolution.

@mvl22, @si-the-pie - can you please let me know where I can get hold of these assets, or even better, perhaps one of you could raise a PR to submit them to the folder I've linked above?  Many thanks.

mvl22:
@si-the-pie Can you add an icon from the set to the API.

PR #279, opened 2018-08-24
Bug fixes

### LINKED ISSUE #278: Default sound heard during LiveRide
Raised by @si-the-pie by email:

I tried Live Ride and instead of hearing a voice I got a default sound. The sound was emitted at the start of the ride and at each junction - so I'm wondering if I have to setup voice or some such. I have previously heard the voice on this device about a year ago.

Ah - now I'm home I have discovered that the media volume was low and the voice is actually there - but the default sound still comes too.  In fact there's a double default sound emitted at the start of the
route.

oliverlockwood:
Closes #278.
Connects to #277, and resolves the UI inconsistency, but does not close (as the true fix requires new icons to be added to the assets in the app).

In addition this PR:
-  Fixes a missing image in the About page (a minor regression introduced by #251)
-  Changes the main fragment transitions to be `Fade` rather than `Slide` - if your phone is low on memory or a bit old, then `Slide` can appear a bit jerky.


oliverlockwood:
@jezhiggins: Please review.
@mvl22: FYI.
@si-the-pie: This should fix the issues you raised by email.

Issue #276, opened 2018-08-22
LiveRide: Directions sometimes confusing


oliverlockwood:
I don't think this is a LiveRide bug, more of a routing issue.

For example, there have been many times that I've approached a crossroads, perhaps going to cross a more major road, and the instruction is given to "turn left".  I duly do so, only to be told to "continue zero metres, get ready to turn right".

@mvl22 is this something you're aware of server side?

If not, or if the fix isn't likely to happen imminently, I might put in a workaround on the app side, e.g. to look ahead and merge instructions if the interval is below 10 metres or so... because this has been annoying me for a while.

jezhiggins:
Oh man, I've thought about this a lot - my name for this is segment melding.  There are some cases when the road is in multiple segments but in reality it's just one continuous road, and those are kind of ok. The other cases, which really can be a drag, are the turn left, go 20 metres, turn right occurrences.  I tried to come up with a scheme that handled both but, with the benefit of some distance, I probably was overthinking it. Have a go.

mvl22:
Can you send some specific example routes if possible. Ideally we want to combine traversal through a junction into the overall direction movement, not individual pieces.

oliverlockwood:
Hi @mvl22, here's a number of specific examples, narrowed down from much longer routes I have in my history.
-  63090075 (10 metres on Bradenham Lane)
-  63090089 (9 metres on Pebble Hill)
-  63090095 (10 metres on High Street, A338)

In each case, you can go to https://www.cyclestreets.net/api/journey.json?plan=balanced&itinerary=<itineraryId>&key=<redacted> to see the raw JSON with the relevant segments; if you load those routes in the app, you'll be able to navigate through the segments and see where there's a problem.

I can find more if necessary but that should give you the general idea.

Watch this space for a PR which demonstrates what the voice notifications sound like in this circumstance.

si-the-pie:
Yes, I agree that the root cause of this is problem lies with the output from the routing engine, which could do more to help.

Look at this route near Blackfriars Bridge in Central London. It starts in Queen Victoria Street and basically **turns left** onto Blackfriar's Bridge, but it is drawn in OSM as a complex wiggle and this is regurgitated in the complex and detailed routing instructions:

https://www.cyclestreets.net/journey/63112198/#fastest

[...]

PR #283, opened 2018-08-24
Add tests to prove the voice notification behaviour for LiveRide

oliverlockwood:
Connects to #276.

This PR does not change functionality.  The only app code changes are refactors to improve testability.  The main point is to add some unit tests which prove what the voice notifications are for a given "confusing" section of a route. 

A subsequent PR can then propose a solution / workaround, and the consequential change to these tests should make that PR easier to review and provide greater confidence in any functional changes we make.


libraries/cyclestreets-view/src/test/java/net/cyclestreets/liveride/LiveRideTest.kt
```diff
@@ -0,0 +1,120 @@
...
+
+        move(-3.33022, 50.92086)
+        verify("Get ready to Turn right")
+        move(-3.33019, 50.92081)
+        verify("Turn right into A38. Continue 15m")
+        move(-3.33019, 50.92081)
+        move(-3.33019, 50.92081)
+        verify("Get ready to Turn left")
+        move(-3.33019, 50.92081)
+        verify("Turn left into Broad Path. Continue 990m")
```
oliverlockwood: These 10 lines above are the interesting part of this PR.

The key point is that it's not until you hit the junction itself and get the "Continue 15m" that you get any clue that you're not going straight across.  If you're cycling quickly on the approach to the junction (if visibility is good), then you could be 30 metres past the turning into Broad Path before you realise your error.

PR #286, opened 2018-08-27
Bike share images to help with #277

si-the-pie:
Connects to #277.

The bikeshare*.png images are just copies of bicycles*.png.


oliverlockwood:
Thanks @si-the-pie

Issue #281, opened 2018-08-24
LiveRide voice oddities


si-the-pie:
The voice refers to itself as: Live (as in 'and Let Die') Ride, rather than: Live (as in 5) Ride.

On a short journey I hear the voice say: "One minutes"

An oddity is hearing: "Getting ready to arreeve eh" which I think is supposed to sound like the french word arrivé.

I want to tap on the screen for it to read out the last instruction again - or say to my phone: Pardon? / Can you say that again!

oliverlockwood:
@si-the-pie - can you please raise the "I want to" as a separate issue with label `enhancement`?

I'll keep this issue to track the oddities.

oliverlockwood:
@si-the-pie can you please review whether the latest build fixes these issues, please?

oliverlockwood:
Scratch that, now it's pronouncing "live-uh" for me on my phone (although correctly on the app).  Another iteration under way.

oliverlockwood:
Closing this as I believe it's now fixed.  @si-the-pie please reopen if you disagree.

PR #287, opened 2018-08-27
LiveRide improvements

### LINKED ISSUE #278: Default sound heard during LiveRide

### LINKED ISSUE #284: LiveRide notification icon is grey square in later Android versions
It's a grey square on my phone.  It looks like it might not be on @si-the-pie's phone - Simon am I right in thinking that you see the CycleStreets logo on your notifications?

I think this is a change in Android that we need to cater for.

oliverlockwood:
This PR:
-  **Closes #278.**
    If you've created a channel with the wrong notification level, you can't change it in subsequent installs of the app.  Rather than force beta users to uninstall, we've just created a separate channel and deprecated / deleted the old one.
-  **Closes #284.**
    The size of small notification icons is such that it makes sense to use a Material design cycle icon, rather than trying to miniaturise a transparent CycleStreets logo to be that small (you wouldn't be able to read "CycleStreets" in the image; in any case, the app name is present as text in the notifications.
    In addition, I've made an enhancement such that when saying "Get ready to turn left" (for example) there'll be a large icon overlay of the relevant direction arrow.
-  **Connects to #281 - and, I hope, fixes it** - but as I've not been able to repro the pronunciation issue, I'll wait for @si-the-pie to test the ensuing build, before closing the issue.
    -  The "1 minutes" case is caught quite early on, and fixed in the visual route display as well as in the voice.
    -  Improved phonetic hack to try to force TTS to say "laɪv" instead of "lɪv" in the context "Live Ride"
    -  Improved phonetic hack for "arrivée" (not "arriverai")
    -  There's separation of what we send to the TTS engine (e.g. the phonetic hacks, above) from what we display on the screen.

Screenshot showing a direction arrow in the LiveRide notification:

screenshot_20180827-122352_01

### DIFF
--- a/libraries/cyclestreets-view/src/test/java/net/cyclestreets/liveride/LiveRideVoiceTest.kt
+++ b/libraries/cyclestreets-view/src/test/java/net/cyclestreets/liveride/LiveRideVoiceTest.kt
@@ -23,7 +23,7 @@ import org.mockito.exceptions.verification.VerificationInOrderFailure
 
 @Config(constants = BuildConfig::class, manifest = Config.NONE, sdk = [27])
 @RunWith(RobolectricTestRunner::class)
-class LiveRideTest {
+class LiveRideVoiceTest {
 
     private lateinit var liveRideState: LiveRideState
     private lateinit var journey: Journey


oliverlockwood:
@jezhiggins @si-the-pie for review, please, so we can fix these issues and proceed towards the release.

jezhiggins:
@oliverlockwood Changes are all fine with me. Haven't got time to look at the merge conflict right now, but loop back around later if you don't get the chance.

oliverlockwood:
Thanks Jez, I've just fixed the merge conflicts.

PR #289, opened 2018-08-28
Upgrade a couple of Gradle plugins

oliverlockwood:
Only minor and patch version bumps, so this should be pretty trivial.


oliverlockwood:
Merging this one as it's trivial.

PR #290, opened 2018-08-28
Broaden LiveRide tests to include Bridges

oliverlockwood:
Also remove some redundant data.


oliverlockwood:
Self-merging as this PR only affects test code, and the checks are green.

PR #291, opened 2018-08-28
Segment melding - staggered crossroads and bridges

oliverlockwood:
Connects to #276.

This PR does three things:
1.  Separate the turn icon derivation from the eventual turn instruction, to allow us to customise instructions as we would in segment meld cases
2.  Melds turn/bear left/right or right/left instructions within 20 metres of one another, including new turn icons to reflect the meld
3.  Melds Bridge segments with the segment that follows, into a single instruction e.g. "Straight on over bridge into ..."

The unit test changes should give a clear indication of the behavioural change that results.

Note that as and when server-side changes are made to address the described issue, we should be able to deprecate this functionality.


libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Segments.kt
```diff
@@ -16,6 +17,11 @@ class Segments : Iterable<Segment> {
     fun first(): Segment.Start { return segments.first as Segment.Start }
     fun last(): Segment.End { return segments.last as Segment.End }
 
+    private val CROSSROAD_MELDS: Map<Pair<Turn, Turn>, Turn> = mapOf(
+        Pair(TURN_LEFT, TURN_RIGHT) to LEFT_RIGHT,
+        Pair(TURN_RIGHT, TURN_LEFT) to RIGHT_LEFT
+    )
+
```
jezhiggins: Neat!

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Segments.kt
```diff
@@ -18,10 +18,10 @@ class Segments : Iterable<Segment> {
...
 
     private val CROSSROAD_MELDS: Map<Pair<Turn, Turn>, Turn> = mapOf(
-        Pair(TURN_LEFT, TURN_RIGHT) to LEFT_RIGHT,
-        Pair(TURN_RIGHT, TURN_LEFT) to RIGHT_LEFT,
-        Pair(BEAR_LEFT, BEAR_RIGHT) to BEAR_LEFT_RIGHT,
-        Pair(BEAR_RIGHT, BEAR_LEFT) to BEAR_RIGHT_LEFT
+        Pair(TURN_LEFT, TURN_RIGHT) to TURN_LEFT_THEN_RIGHT,
+        Pair(TURN_RIGHT, TURN_LEFT) to TURN_RIGHT_THEN_LEFT,
+        Pair(BEAR_LEFT, BEAR_RIGHT) to BEAR_LEFT_THEN_RIGHT,
+        Pair(BEAR_RIGHT, BEAR_LEFT) to BEAR_RIGHT_THEN_LEFT
```
jezhiggins: Presumably we can extend this to the cross product of turn_* and bear_* if we're feeling energetic?

oliverlockwood: Yep... I'll probably do this as a follow up when I'm ready to face another round of image editing in GIMP :smile:

PR #292, opened 2018-08-28
Give up trying to say "Live Ride"

oliverlockwood:
Connects to #281.

I've lost count of how many attempts I've had at spelling "Live" phonetically.

On my phone, "Live" works as intended; on more recent Android versions, it's apparently pronounced "Liv".

The closest I've got phonetically is "Lyev" which sounds half way between "Live" and "Lie-ev".

Eventually I came to the conclusion that it really doesn't add very much, so we're better off removing those two words.


si-the-pie:
The second definition from [Google's dictionary](https://www.google.co.uk/search?q=tomoz&oq=tomoz&aqs=chrome..69i57.1151j0j7&sourceid=chrome&ie=UTF-8#dobs=live) uses: lʌɪv

oliverlockwood:
@si-the-pie I thought I had already tried this, but just to make sure, I tried again.  It vocalised this as "L V".

oliverlockwood:
I've now also extended this PR slightly to:
-  refactor `LiveRideService` into Kotlin, and
-  as soon as the TTS engine is initialised, force-send a location update (otherwise it won't tell you where you're going until you've moved at least a few metres, which is a little bit frustrating when you're setting off on the bike, and very annoying indeed when you're testing inside an office building and have no GPS signal 😄)

PR #300, opened 2018-08-30
Fix route finding failures for single-segment routes

### LINKED ISSUE #298: Route finding failed
I've been occasionally getting the 'Route finding failed' message after setting waypoints by tapping. Usually if I re-request a route then one does appear. But this time it did not, and below is the screen grab for Thoday Street, Cambridge.

That route generates this request (api key hidden):

`https://www.cyclestreets.net/api/journey.json?plan=balanced&itinerarypoints=0.1478084921836853%2C52.2000287382715%7C0.14693140983581543%2C52.198634652097525&speed=16&key=`

(incidentally it is only necessary to send co-ordinates rounded to five places of decimals, which is a lot easier to read)

and generates a route like this one: https://www.cyclestreets.net/journey/63122347/#balanced

so I don't think the error is from the CycleStreets server. I have been able to reproduce the problem by removing the waypoints and resetting them. Sometimes I get a route, sometimes not.

screenshot_20180829-204753_cyclestreets

oliverlockwood:
Closes #298.

This PR also
-  improves the logging we do around route finding failures
-  introduces OkHttp's `HttpLoggingInterceptor`, but leaves it disabled by default since at the moment there doesn't seem to be a way to stop it from logging the API key, which I'd rather not do (see https://github.com/square/okhttp/issues/4244).

### DIFF
--- a/libraries/cyclestreets-core/build.gradle
+++ b/libraries/cyclestreets-core/build.gradle
@@ -21,6 +21,7 @@ dependencies {
     exclude group: 'stax', module: 'stax'
   }
   implementation 'com.squareup.retrofit2:converter-jackson:2.4.0'
+  implementation 'com.squareup.okhttp3:logging-interceptor:3.11.0'
   implementation 'de.grundid.opendatalab:geojson-jackson:1.8'
 
   // Jackson is already included transitively from multiple sources; we specify explicit versions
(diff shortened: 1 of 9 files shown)


oliverlockwood:
Actually, I'm confident enough having tested this small fix, that I'm going to self-merge.  @jezhiggins please do review when you have the chance, and I'll address any comments retrospectively.

Issue #42, opened 2011-11-27
Lock screen navigation mode


smsm1:
To prevent accidentally coming out of the app and losing your navigation position whilst moving, it would be good if there was something similar to the Google Navigation app whereby the app is shown in the lock screen, with the navigation updating as you move along.

oliverlockwood:
On LiveRide, there's a lock button in the top right to keep the screen on (which I think is the implementation Jez was suggesting in his comments above).  @mvl22 Does this cover the requirements here?

mvl22:
I think we should do the same as whatever other navigation apps like Google do. I would assume there is some kind of setting perhaps. Certainly the default should be that locking is not on while LiveRide is running.

jezhiggins:
The whole landscpe around this has changed hugely since I first wrote the live ride stuff.  At the time, you could either keep the screen on (which is what the lock does) or let it turn itself off.  There was no standard way to turn it back on when something exciting was happening.  

Similarly, while back in 2011/12 some apps could display fancy lock screen widgets, that required phone specific code - there was no standard API to support it.  However, the lock screen widget have largely been killed by the notifications which, in these modern times, can include graphics and controls, rather than being the simple text strings they were.

So, it's all up for grabs really.

mvl22:
I think it just needs a bit of text underneath that says either "Lock screen" or "Screen locked".

Any chance would also get the attribution at the bottom not overlaid, but put in the bottom-right corner?

oliverlockwood:
Alternatively @mvl22 @si-the-pie perhaps you could look at https://material.io/tools/icons and see if you can come up with some better icons?

@mvl22 Please raise a separate issue to cover the map attribution location.

mvl22:
How about `screen_lock_portrait`, which is gray by default but green when enabled?

Will add a separate issue - sorry.

si-the-pie:
I've had a look thru the available icons in the material design and none of them strike me as an ideal fit for the stay awake effect.

I don't like the `screen_lock_portrait` icon because that is I think intended to stop rotating the screen. I already said that the use of a lock to stay on is confusing because it contradicts with 'Lock screen'. So I think that would end up as a right mess.

Thinking about the actual use case, during a LiveRide the screen should definitely stay awake while the current location lies somewhere on the route.

But keeping the screen turned on is a security issue, so really what is needed is a longer timeout. The kind of timeout longer than a wait at a set of traffic signals.

[...]

PR #303, opened 2018-08-30
Change the lock-screen-on button

oliverlockwood:
Connects to #42.

## In the previous public release, the LiveRide screen stay-alive button looked like this:
screen shot 2018-07-23 at 09 40 57
screen shot 2018-07-23 at 09 41 05

## In the current beta, it looks like this:
screen shot 2018-07-23 at 09 35 54
screen shot 2018-07-23 at 09 36 02

## After this PR:
screen shot 2018-08-30 at 11 29 44
screen shot 2018-08-30 at 11 29 51


oliverlockwood:
@si-the-pie @mvl22 please review this with the context of wanting to get the new release 3.6 out!

si-the-pie:
The green coloured outline padlock is a subtle distinction from the gray outline padlock, especially outdoors.
Can I suggest that the eye image (called `visibility` in the material icons listing) is used instead of the green padlock.
Colour it green if you want. This should be an improvement on the current position.

Another issue is that it takes two taps to change the initial state. The first tap is detected as the button flickers but does not change state.

(As tested with version 1505.)

oliverlockwood:
@si-the-pie are you saying it should be the `lock_open` image when disabled, and the `visibility` icon when enabled?  To my mind that would make things even less clear.

@mvl22 ?

si-the-pie:
The button is really there to turn off screen timeout. [I originally suggested](https://github.com/cyclestreets/android/issues/42#issuecomment-416972762) that a tool tip briefly appears to explain the new state, i.e. 'Screen will stay on' or 'Screen will timeout'.

Can you tell us whether it is possible to have such tool tip?

If a tool tip is available then that would help unravel what this button is doing. The button must have two quite distinct images so that it can easily be discerned when looking at the screen from a distance.

Looking through the material icons how about:

- `timer` for the _Screen will timeout_
- `visibility` for _Screen will stay on_

These are quite distinct, but need the tooltip to avoid confusion.

I am making these suggestions within the constraints of a workaround rather than [another design](https://github.com/cyclestreets/android/issues/42#issuecomment-417097332).

mvl22:
> eye image

I agree with Oliver that this is even less clear.

The proper solution here is a label, as I don't think any icon on its own will be enough. As Oliver suggested, I think that should go in the next release rather than hold this up here. There is a *lot* going into this 3.6 release and I really want to get it out the door.

PR #304, opened 2018-08-30
Fix recursion bug; add a logging interceptor

oliverlockwood:
Adding in a custom logging interceptor so we don't log the API key.

Also fixing a recursion bug accidentally introduced by a previous PR - my bad.


libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/ApiClient.kt
```diff
@@ -113,16 +113,16 @@ object ApiClient : CycleStreetsApi {
         return delegate.register(username, password, name, email)
     }
     override fun getPOICategories(iconSize: Int): POICategories {
-        return getPOICategories(iconSize)
+        return delegate.getPOICategories(iconSize)
```
oliverlockwood: It's a while since I got a `StackOverflowException`.  The shame 😬

oliverlockwood:
Merging this because the stack overflow bug is critical.

PR #293, opened 2018-08-29
Replan keeps remaining waypoints

oliverlockwood:
Closes #106.

This PR includes:
-  refactoring a couple of classes into Kotlin
-  refactoring the ApiClient to aid testability, so we can stub the API without a full Wiremock stub (I'm not especially keen on the layers of obfuscation here, but I think it's OK for now and I can look at improving subsequently)
-  adding (failing) unit tests to specify our desired behaviour
-  changing the `ReplanFromHere` class to keep intermediate waypoints, closing the linked issue.

@jezhiggins would you mind reviewing this, please?

Issue #257, opened 2018-08-10
New screenshots for page on CycleStreets website


mvl22:
This issue is to get some new screenshots to replace the ancient ones at:
https://www.cyclestreets.net/mobile/android/

A reminder that a good screenshot contains places that are more likely to be recognisable to a larger number of people, e.g. central London's Downing Street or the Bullring in area in Birmingham, rather than an obscure village. For screenshots from the Photomap, good, positive images, like people without helmets/hi-viz of various genders using a cycle superhighway, would be good.

oliverlockwood:
I think:
1. We should get new screenshots
2.  Perhaps you'd like to do the honours?
3.  As mentioned in the other issue on the subject, we should update the images on the play store but I believe *the cyclestreets page should refer to this for its images*. Otherwise we have two sources of truth to maintain, no?

oliverlockwood:
Marking this as milestone 3.7 (since we'll do it *after* we're ready to release 3.6, even though it relates to release 3.6).

mvl22:
As I don't currently have a modern-enough Android, I'd be grateful if you could make and e-mail me some screenshots, so that I can update the page on the CycleStreets website.

See note above about recognisable locations.

oliverlockwood:
@mvl22 did you see my comment above about trying to avoid having two separate sources of truth for the images?
>we should update the images on the Play store but I believe the Cyclestreets web page should refer to this for its images, to avoid duplication

mvl22:
Yes, sorry was actually just editing my comment accordingly...! Here's what I have just rewritten:

Yes, I'll use ones that get submitted to the Android Play store. See note above about recognisable locations.

However, I suggest shifting this down to 3.6, as the release should be accompanied by updated screenshots. The ones there are old and have random unrecognisable places.

oliverlockwood:
@mvl22 now that you've got a load of relevant screenshots, and I've updated the [Play Store entry](https://play.google.com/store/apps/details?id=net.cyclestreets) with them, can you please get https://www.cyclestreets.net/mobile/android/ updated accordingly ASAP?

oliverlockwood:
@mvl22 has done this now.  Closing the issue accordingly.

PR #311, opened 2018-08-31
Fix regression (no photos in photomap!) caused by inconsistent argument ordering

oliverlockwood:
PR #293 refactored the API client.

This introduced a bug, because the arguments for various APIs (west, south, east, north) were ordered inconsistently.

This PR fixes that bug and makes the argument ordering consistent across the API and its use cases, in a way that matches the underlying API itself.

(Bug was spotted while gathering screenshots for #257.)


mvl22:
Thanks, well spotted.

WSEN is the order we use throughout the code for a bbox, basically matching the standard established by Leaflet.js.

oliverlockwood:
Indeed, that's the convention I've established in the app code as well.

PR #314, opened 2018-08-31
Explanatory messages for Live Ride screen lock button

oliverlockwood:
Relates to discussions in #42 and #303.

This PR does not change the images that are used (since we have yet to come to an agreement around that) but it does add an explanatory message (slightly tweaked from @si-the-pie's original suggestions).  It also fixes the you-need-to-click-twice-the-first-time issue.

I hope this is sufficient for the 3.6 release.

@si-the-pie @mvl22 when you've discussed between yourselves and agreed what the icons should be for the subsequent 3.7 release, please raise a new issue accordingly (or just a PR - it should be pretty obvious what the trivial changes would be to adjust which icons are chosen).

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LockScreenOnOverlay.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LockScreenOnOverlay.java
@@ -1,5 +1,6 @@
 package net.cyclestreets.views.overlay;
 
+import android.widget.Toast;
 import net.cyclestreets.util.Theme;
 import net.cyclestreets.view.R;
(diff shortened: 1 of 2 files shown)


mvl22:
Could you screenshot this?

oliverlockwood:
Screenshots (from the emulator) as requested are below:
screen shot 2018-08-31 at 13 26 20
screen shot 2018-08-31 at 13 26 15

mvl22:
OK, that Toast Message is a good solution. Ideally make it slightly higher as it's visually being connected to the bottom-left.

oliverlockwood:
You can't control where the toast message appears, I'm afraid.  I'll merge it as-is.

PR #315, opened 2018-08-31
Version control the Play Store listing assets

oliverlockwood:
This PR makes more extensive use of the https://github.com/Triple-T/gradle-play-publisher plugin - not only to publish our builds, but also to update our Store listing.

Note that I have run the Gradle task already, so our store listing now includes the updated screenshots.  (It apparently takes a few hours for these changes to be reflected in what users see in the store).


oliverlockwood:
Merging this to get on with the release tagging etc.

PR #320, opened 2018-09-04
Highlight active segment on elevation graph

oliverlockwood:
So that it's made possible to see *where* on a route a steep section might fall, for example.

### Resulting screenshot:
screen shot 2018-09-04 at 15 08 29


oliverlockwood:
@jezhiggins would you mind reviewing, please?

PR #321, opened 2018-09-04
Prevent liveride restart on wake

jezhiggins:
There might be an element of device/platform dependency here but, for me at least, LiveRide restarts everytime on wake and, often, seems to be killed on pause. Hopefully this will prevent that - need to go out and about to verify.


libraries/cyclestreets-view/src/main/java/net/cyclestreets/LiveRideActivity.kt
```diff
@@ -19,15 +26,39 @@ import net.cyclestreets.util.hasPermission
...
+
+    public override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        val intent = Intent(this, LiveRideService::class.java)
+        this.bindService(intent, this, Context.BIND_AUTO_CREATE)
+    }
+
+    public override fun onDestroy() {
+        this.unbindService(this)
```
oliverlockwood: Do we need to call
```kotlin
liveride?.stopRiding
```
here?  Previously we did (see https://github.com/cyclestreets/android/pull/321/files#diff-78f316180ff253b501bc22d4a1cd8799L76)

oliverlockwood:
For what it's worth, my personal experience (I always sleep the phone in my back pocket while riding) is that the directions continue to be given (i.e. the service is not killed), but if I wake the phone (e.g. to look at how the map relates to the verbal instructions I've just heard) then it does restart the service (going through the whole route description, length, duration spiel etc.) which is quite annoying.

So thanks for working on improving this!

I've made one inline comment.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/LiveRideActivity.kt
jezhiggins: Yes.  The service will stop itself when it's destroyed, which is triggered by unbinding, but it's probably better to shut it down first.

jezhiggins:
Had a couple of rides around with this, and happy it's all working as it should.

PR #319, opened 2018-09-04
Support location urls

oliverlockwood:
Connects to #91.

This PR enables handling of location URLs by the CycleStreets app.

I intend to have follow-up PRs look at:
1.  handling Geo intents, and other mapping locations
2.  allowing users to share Photomap entries

PR #322, opened 2018-09-05
Allow HTTPS intents for cycle.st URLs

oliverlockwood:
Addresses @mvl22's post-merge comment https://github.com/cyclestreets/android/commit/f0441828cdef6d23d2436459e08f07d375103477#commitcomment-30390774 following PR #319

>Just to flag up that cycle.st is about to get HTTPS at last - for some reason this was accidentally missed when we shifted our infrastructure to HTTPS. So please do include this in the list of intents.


oliverlockwood:
Self-merging this trivial tweak.

PR #324, opened 2018-09-05
Extra replan waypoint should be temporary

### LINKED ISSUE #323: Waypoint replanning doesn't work quite as intended
Follows on from #106 and its related PRs.

### The intention:
-  Keep whichever remaining *planned* waypoints are appropriate (i.e. the ones that have not yet been passed)
-  Add a starting waypoint of current location.

### The problem:
-  This works fine if *after the first replan, you get right back on track*.
-  If you don't, then *subsequent replans include the location of the last replan as an additional waypoint*.  This is bad, as the route goes more and more mental as you try to shortcut towards the last intentional waypoint!

### Screenshot of the problem:
)

### The proposed solution:
-  Define a type for waypoints - planned, or ad-hoc.  Ad-hoc ones are only generated by on-the-fly replanning, and *are not kept in any subsequent replans*.

oliverlockwood:
Closes #323.

As usual at the moment there's a bunch of Kotlin refactors.  I've tried to keep these as separate commits; so the interesting commits to review are named "Add some failing tests" and "Don't include ephemeral waypoints when replanning".

@jezhiggins would you please review?


oliverlockwood:
With the unit tests in place I'm pretty confident this should do what's needed.

I'm merging this now so I can easily test on the way home (I don't have a USB cable with me here in the office).

Please do review when you have the chance and I will address any comments retrospectively.

oliverlockwood:
Tests on the way home were very successful. I'm confident that this has resolved the problem.

PR #326, opened 2018-09-06
Find place dialog improvements

### LINKED ISSUE #78: Find Place popup covers Find button.
This is a problem on slow connections:
1. Open _Find Place_
2. Type your search
3. Nothing happens because your connection is slow
4. Try to click the _Find_ button
5. Search-as-you-type popup appears
6. click something random in the popup by mistake.

Suggested solution
Copy Google maps - Move the find button so that it replaces enter on the on screen keyboard, and use a magnifying glass symbol.

oliverlockwood:
Closes #78.

Best to review commit-by-commit to separate out the Kotlin migration.

This PR doesn't change the Find button.  Instead it:
-  makes the pop-up keyboard have a "search" action (with magnifying glass icon, which invokes Find directly) instead of an "enter" action (which pointlessly adds a newline inside the text box).
-  reduces the max width of the drop-down to 90% of the width of the text box itself, so it doesn't obscure the Find button.

## Screenshot before
screen shot 2018-09-06 at 14 36 34

## Screenshot after
screen shot 2018-09-06 at 14 59 27


libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleMapFragment.kt
```diff
@@ -135,7 +135,7 @@ open class CycleMapFragment : Fragment(), Undoable {
     }
 
     private fun launchFindDialog() {
-        FindPlace.launch(activity, map!!.boundingBox) { place ->
+        FindPlace.launch(context!!, map!!.boundingBox) { place ->
```
jezhiggins: Do we need the !! here?  Can context or map actually null?

oliverlockwood: @jezhiggins good question; but yes, we do need these.

-  Previously, `FindPlace` was in Java and the `launch` method took a (implicitly nullable) `Context` object.  If it was null (unlikely as that is, clearly) then that would have been a runtime failure.
-  Now, `FindPlace` is in Kotlin, and to avoid peppering that class with `!!` throughout, the `launch` method takes a `Context` object (explicitly non-nullable, as opposed to `Context?` which is explicitly nullable).
-  Since the `Fragment.getContext()` method (and indeed `Fragment.getActivity()` similarly) is annotated as `@Nullable`, then we have to assert here at compile time.

In the case of the `map` property, look at the history of PR #273 - I originally had this as a `lateinit var` (so non-nullable) but found that this caused unnecessary assertion crashes in certain scenarios, so changed it to be nullable.

oliverlockwood:
Merging as approved.

libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleMapFragment.kt
jezhiggins: Ah, gotcha.  So we're moving the potential failure site up out of FindPlace to the callsite.  If can, in the future, assert that these things are non-null, the linter will tell and we can clean them up then.

PR #328, opened 2018-09-07
Waypoints on LiveRide

jezhiggins:


oliverlockwood:
@jezhiggins this looks like another merge-confused PR, I'm afraid.  I'm pretty sure this PR doesn't actually change 101 files 😄

Can I suggest creating a fresh branch off an up-to-date `master`, and then cherry-picking the relevant commits?  Something like this (assuming `origin` is set to the `cyclestreets` root, not your fork):
```
git checkout master
git pull origin master
git branch -D waypoints-on-liveride
git checkout -b waypoints-on-liveride
git cherry-pick d62c424 2258722 546a410 18394dd
git push origin -f
```

jezhiggins:
Yea, that's me believing GitHub when it says "here's how you sort out this merge conflict".  (Leaving aside the fact I branched off another branch to begin with so have already gone round this loop once already :)

jezhiggins:
That's better

oliverlockwood:
😅

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,170 @@
...
+        redWisp = ResourcesCompat.getDrawable(res, R.drawable.red_wisp, null)
+
+        shareDrawable = IconicsDrawable(context)
+                .icon(GoogleMaterial.Icon.gmd_share)
+                .color(Theme.lowlightColorInverse(context))
+                .sizeDp(24)
+        commentDrawable = IconicsDrawable(context)
+                .icon(GoogleMaterial.Icon.gmd_comment)
+                .color(Theme.lowlightColorInverse(context))
+                .sizeDp(24)
```
oliverlockwood: I don't think we need these two drawables in this class.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,170 @@
...
+
+        waymarkers = ArrayList()
+    }
+
+    private fun setWaypoints(waypoints: Waypoints) {
+        resetWaypoints()
+
+        for (waypoint in waypoints) {
+            addWaypoint(waypoint)
+        }
```
oliverlockwood: This `for` loop could be replaced with:
```kotlin
        waypoints.forEach { wp -> addWaypoint(wp) }
```

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    fun waymarkersCount(): Int {
+        return waymarkers.size
+    }
+
+    fun waypoints(): Waypoints {
+        val p = ArrayList<IGeoPoint>()
+        for (o in waymarkers)
+            p.add(o.point)
+        return Waypoints(p)
```
oliverlockwood: Since we're now in Kotlin, this whole method could be a one-liner:
```kotlin
        return Waypoints(waymarkers.map { wm -> wm.point })
```

🎉

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,170 @@
...
+                waymarkers.add(addMarker(prevFinished, "waypoint", orangeWisp))
+                waymarkers.add(addMarker(point, "finish", redWisp))
+            }
+        }
+    }
+
+    fun removeWaypoint() {
+        when (waymarkersCount()) {
+            0 -> {
+            }
```
oliverlockwood: make this a oneline no-op rather than splitting over two lines?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,170 @@
...
+                waymarkers.add(addMarker(prevFinished, "finish", redWisp))
+            }
+        }
+    }
+
+    private fun addMarker(point: IGeoPoint, label: String, icon: Drawable?): OverlayItem {
+        val marker = OverlayItem(label, label, GeoPoint(point.latitude, point.longitude))
+        marker.setMarker(icon)
+        marker.markerHotspot = OverlayItem.HotspotPlace.BOTTOM_CENTER
+        return marker
```
oliverlockwood: Not essential, but if you want to make this more concise, you could change it to:

```kotlin
        return OverlayItem(label, label, GeoPoint(point.latitude, point.longitude)).apply {
            setMarker(icon)
            markerHotspot = OverlayItem.HotspotPlace.BOTTOM_CENTER
        }
```

jezhiggins:
Thanks - this is really helpful stuff. I need get a bit more up to speed on Kotlin - these kind of transformations are the kind of thing I'd do without even thinking in pretty much any other language (even Java8 :open_mouth:), but I'm so used to Android having become a legacy platform (which is weird in itself, right?) that it didn't even occur. Do stay vigilant while I get acclimatized.

oliverlockwood:
My pleasure, glad to help!

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        waymarks.addWaypoint(point)
+        controller.pushUndo(this)
+        tapState = tapState.next(waypointsCount())
+        mapView.invalidate()
+    }
+
+    ////////////////////////////////////
+    private enum class TapToRoute private constructor(val waypointingInProgress: Boolean, val actionDescription: String) {
```
oliverlockwood: I wonder about moving this enum out into its own file in the same package - as an `internal enum` rather than a `private enum`.  What do you think?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+        ALL_DONE(false, "");
+
+        fun previous(count: Int): TapToRoute {
+            val previous: TapToRoute
+            when (this) {
+                WAITING_FOR_START, WAITING_FOR_SECOND, ALL_DONE -> previous = WAITING_FOR_START
+                WAITING_FOR_NEXT -> previous = if (count == 1) WAITING_FOR_SECOND else WAITING_FOR_NEXT
+                WAITING_TO_ROUTE -> previous = WAITING_FOR_NEXT
+            }
+            Log.d(TAG, "Moving to previous TapToRoute state=" + previous.name + " with waypoints=" + count)
```
oliverlockwood: In Kotlin we can use string interpolation:
```kotlin
            Log.d(TAG, "Moving to previous TapToRoute state=${previous.name} with $count waypoints")
```

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+
+        fun next(count: Int): TapToRoute {
+            val next: TapToRoute
+            when (this) {
+                WAITING_FOR_START -> next = WAITING_FOR_SECOND
+                WAITING_FOR_SECOND -> next = WAITING_FOR_NEXT
+                WAITING_FOR_NEXT -> next = if (count == MAX_WAYPOINTS) WAITING_TO_ROUTE else WAITING_FOR_NEXT
+                WAITING_TO_ROUTE, ALL_DONE -> next = ALL_DONE
+            }
+            Log.d(TAG, "Moving to next TapToRoute state=" + next.name + " with waypoints=" + count)
```
oliverlockwood: As above - string interpolation

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+        private val TAG = Logging.getTag(TapToRouteOverlay::class.java)
+        private val REPLAN_MENU_IDS = intArrayOf(R.string.route_menu_change_replan_quietest, R.string.route_menu_change_replan_balanced, R.string.route_menu_change_replan_fastest, R.string.route_menu_change_replan_shortest)
+        private val REPLAN_MENU_PLANS = object : HashMap<Int, String>() {
+            init {
+                put(R.string.route_menu_change_replan_quietest, RoutePlans.PLAN_QUIETEST)
+                put(R.string.route_menu_change_replan_balanced, RoutePlans.PLAN_BALANCED)
+                put(R.string.route_menu_change_replan_fastest, RoutePlans.PLAN_FASTEST)
+                put(R.string.route_menu_change_replan_shortest, RoutePlans.PLAN_SHORTEST)
+            }
+        }
```
oliverlockwood: Kotlin allows us a nicer way of statically initialising maps:
```kotlin
        private val REPLAN_MENU_PLANS = mapOf(
            R.string.route_menu_change_replan_quietest to RoutePlans.PLAN_QUIETEST,
            R.string.route_menu_change_replan_balanced to RoutePlans.PLAN_BALANCED,
            R.string.route_menu_change_replan_fastest to RoutePlans.PLAN_FASTEST,
            R.string.route_menu_change_replan_shortest to RoutePlans.PLAN_SHORTEST
        )
```

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+        setRoute(journey.isEmpty())
+    }
+
+    override fun onResetJourney() {
+        resetRoute()
+    }
+
+    companion object {
+        private val TAG = Logging.getTag(TapToRouteOverlay::class.java)
+        private val REPLAN_MENU_IDS = intArrayOf(R.string.route_menu_change_replan_quietest, R.string.route_menu_change_replan_balanced, R.string.route_menu_change_replan_fastest, R.string.route_menu_change_replan_shortest)
```
oliverlockwood: I think `arrayOf` would be fine - the compiler will figure out the type

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.osmdroid.views.overlay.Overlay
+import org.osmdroid.views.overlay.OverlayItem
+
+import java.util.ArrayList
+
+class WaymarkOverlay(private val mapView: CycleMapView) : Overlay(), PauseResumeListener, Route.Listener {
+
+    private val greenWisp: Drawable?
+    private val orangeWisp: Drawable?
+    private val redWisp: Drawable?
```
oliverlockwood: We could consider making these non-nullable and forcing `!!` where they're initialised.  (If we can't find the wisp, it's a big problem, after all...)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    init {
+        val res = mapView.context.resources
+        val makeWisp = { drawable: Int -> ResourcesCompat.getDrawable(res, drawable, null) }
+
+        greenWisp = makeWisp(R.drawable.green_wisp)
+        orangeWisp = makeWisp(R.drawable.orange_wisp)
+        redWisp = makeWisp(R.drawable.red_wisp)
+
+        waymarkers = ArrayList()
```
oliverlockwood: I suggest inlining this initialisation to where `waymarkers` is defined

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,159 @@
...
+        return Waypoints(waymarkers.map { wp -> wp.point })
+    }
+
+    fun finish(): IGeoPoint {
+        return waymarkers.last().point
+    }
+
+    fun addWaypoint(point: IGeoPoint?) {
+        if (point == null)
+            return
```
oliverlockwood: I'm trying to figure out under what circumstances `point` would be `null` and I can't think of any.

Certainly, where it's now called from `tapAction()` in the other `Overlay`, the parameter passed in is non-nullable (so if it is null we'll fail at a higher level than this.)

Maybe we could just make this non-nullable and then see what happens when we play with the app...?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,159 @@
...
+            setMarker(icon)
+            markerHotspot = OverlayItem.HotspotPlace.BOTTOM_CENTER
+        }
+    }
+
+    ////////////////////////////////////////////
+    override fun draw(canvas: Canvas, mapView: MapView, shadow: Boolean) {
+        val projection = mapView.projection
+
+        waymarkers.forEach { wp -> drawMarker(canvas, projection, wp) }
```
oliverlockwood: Maybe inline `mapView.projection` in here?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -0,0 +1,159 @@
...
+    override fun onResume(prefs: SharedPreferences) {
+        Route.registerListener(this)
+    }
+
+    override fun onPause(prefs: Editor) {
+        Route.unregisterListener(this)
+    }
+
+    override fun onNewJourney(journey: Journey, waypoints: Waypoints) {
+        setWaypoints(waypoints)
```
oliverlockwood: I wonder whether it would be clearer just to inline:
```kotlin
    waymarkers.clear()
    waypoints.forEach { wp -> addWaypoint(wp) }
```

and similarly just the first of those lines in the `onResetJourney()` callback below.

I can't see any other uses of the `setWaypoints()` and `resetWaypoints()` methods.

jezhiggins: I did think about that, but I decided the names were helpful.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
jezhiggins: mapView.projection returns a new object on each invocation, and object allocation when drawing is generally considered a bad thing. (Or at least was - whether it's still an issue on modern phones who knows?)  Hoisting the call out of the forEach ensures we only call it once.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
jezhiggins: I wondered too.  I tried to remembered if that's there because it did happen, or due to an abundance of over caution.  I'll give it a go.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
jezhiggins: Doh

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+        routingInfoRect = routeView.findViewById(R.id.routing_info_rect)
+        routingInfoRect.setOnClickListener { _ -> onRouteNow(waypoints()) }
+
+        restartButton = routeView.findViewById(R.id.restartbutton)
+        restartButton.setOnClickListener { _ -> tapRestart() }
+
+        routeNowIcon = routeView.findViewById(R.id.route_now_icon)
+
+        lowlightColour = Theme.lowlightColor(context) or -0x1000000
+        highlightColour = Theme.highlightColor(context) or -0x1000000
```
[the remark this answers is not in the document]
jezhiggins: Android Studio made that change.  I noticed it, but assumed it was correct and the colours are working.  If it looks wrong to you too, then I'll switch back.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+import org.osmdroid.views.MapView
+import org.osmdroid.views.overlay.Overlay
+
+import android.content.Context
+import android.content.Intent
+import android.content.SharedPreferences
+import android.content.SharedPreferences.Editor
+import android.graphics.Canvas
+import android.graphics.drawable.Drawable
+import android.location.Location
```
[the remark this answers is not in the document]
jezhiggins: Indeed.

jezhiggins:
Huh - just found a problem.  The pause/resume behaviour is broken.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
oliverlockwood: Ah OK, I didn't know that!

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -0,0 +1,352 @@
...
+
+    private val waymarks = WaymarkOverlay(mapView)
+    private val controller = OverlayHelper(mapView).controller()
+
+    private var tapState = TapToRoute.WAITING_FOR_START
+
+    init {
+        mapView.overlayPushTop(waymarks)
+
+        context = mapView.context
```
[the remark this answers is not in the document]
jezhiggins: Yep, and I can use it consistently instead of mapView.context throughout the methods too

oliverlockwood:
@jezhiggins what's the latest on this PR?

jezhiggins:
Good to go, I reckon, unless you think otherwise.

oliverlockwood:
Cool.  I've just removed `[Work In Progress]` from the title so we don't end up with silly text in the squash-commit on `master` 😃

PR #329, opened 2018-09-07
Address deprecation warnings

oliverlockwood:
Fixes a bunch of deprecation warnings.

Also reduces the number of warnings we get for `ProgressDialog` to one, so it's less spammy.


libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
```diff
@@ -135,24 +138,44 @@ public static int replanDistance() {
...
-    final String resize = uploadSize();
-    if ("640px".equals(resize))
+    if ("640px".equals(uploadSize()))
       return true;
-    if ("big".equals(resize))
+    if ("big".equals(uploadSize()))
       return false;
 
+    return !onFastConnection();
+  }
```
oliverlockwood: @jezhiggins It would be a bit challenging to test all this stuff.

So I'd appreciate a careful second pair of eyes sanity checking through code read, please!

PR #330, opened 2018-09-07
Shift map attribution right in LiveRide

### LINKED ISSUE #296: Map attribution location
As per the comment on this screenshot:
https://github.com/cyclestreets/android/issues/42#issuecomment-416972762

Any chance would also get the attribution at the bottom not overlaid, but put in the bottom-right corner?

oliverlockwood:
Closes #296.

Attribution is still centred in the middle of the screen usually, but in LiveRide we move it to be centred three-quarters of the way across.  See screenshot below.

screen shot 2018-09-07 at 13 09 15


mvl22:
Thanks.

The new design avoids this problem entirely by putting the OSM attribution in the main menu.

PR #331, opened 2018-09-07
Describe target street when saying 'Get ready to <turn>'

### LINKED ISSUE #26: Better text directions, e.g.: "Cycle for 100 metres then turn left into York Street"
Better text directions, e.g.: "Cycle for 20 metres then turn left", and "Turn left after ..."

This is a problem at the API end, not the app end but it may have knock-on consequences for app design to make better use of such data.

oliverlockwood:
Closes #26.

@mvl22 @jezhiggins are you happy with this change?


jezhiggins:
Absolutely

PR #335, opened 2018-09-21
Fix DB upgrade test when running against Android 9 emulator

oliverlockwood:
A weirdness that I noticed when trying to reproduce a separate issue.
This was pretty painful to figure out!

@jezhiggins can you review, please?

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/content/DatabaseHelper.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/content/DatabaseHelper.java
@@ -9,6 +9,8 @@ import android.provider.BaseColumns;
 import android.util.Log;
 import net.cyclestreets.api.client.JourneyStringTransformerKt;
 import net.cyclestreets.util.Logging;
+import net.cyclestreets.view.BuildConfig;
+
 import org.osmdroid.api.IGeoPoint;
 import org.osmdroid.util.GeoPoint;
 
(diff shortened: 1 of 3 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/content/DatabaseHelper.java
```diff
@@ -46,6 +48,16 @@
...
+
+    // From Android 9 (API version 28, SQLite version 3.22.0) Compatibility WAL is enabled by default.
+    // This breaks our testing mechanism of pasting over a new version of the database, so disable
+    // it when we're in debug mode.  See:
+    // -  https://source.android.com/devices/tech/perf/compatibility-wal
+    // -  https://www.sqlite.org/wal.html
+    // -  https://www.sqlite.org/tempfiles.html
+    if (BuildConfig.DEBUG) {
+      this.setWriteAheadLoggingEnabled(false);
+    }
```
jezhiggins: Is this only required for the tests?  If so, is it possible to move it into the test fixture?

oliverlockwood: I tried that.  Unfortunately, what that means is that when the app starts up (before this point) and initialises the DB, somehow SQLite tracks the state that it's running in WAL mode. So when you then replace the DB (even if WAL mode is disabled in the tests, and regardless of whether you leave or delete the `-shm` and `-wal` files), then SQLite fails to get a readable or writeable database, claiming it's already locked. 😞 

So unfortunately, in order to keep the ability to test in this way (which was incredibly useful when making sure DB migration didn't break anything) we do have to have this tweak to our code.  However, since it's wrapped in the if test, it'll be compiled out in release mode, so it's just pollution to our source code and not the final artifact.

jezhiggins: I did wonder if there was a sequencing issue.

PR #341, opened 2018-10-15
Fix crashes reported in developer console

oliverlockwood:
Code a little more defensively here and there.


oliverlockwood:
@jezhiggins please review when you have the chance.  Thanks.

PR #344, opened 2018-10-19
Fix NPE when hitting Back

oliverlockwood:
Regression caused by #343.

I believe this may be a window condition on upgrading to the new version.
I don't want to lose the repro scenario from my phone, so I'm pushing a change through which I believe should fix things.


oliverlockwood:
@jezhiggins FYI - I'm gonna self-merge and then test this once the build gets published.

jezhiggins:
:+1:

PR #345, opened 2018-10-19
[Work in Progress] Manage offline maps inside main app

oliverlockwood:
Very much a WIP.  For now I just want an easy way to view the diff so I can remember what I was doing when I pick this up every so often.


oliverlockwood:
Closed in favour of #417.

PR #355, opened 2019-10-31
padding updated

rachitchillar:


oliverlockwood:
@rachitchillar perhaps you can provide some justification and context for this proposed change?

If not I will reject it, for reasons that I trust are fairly obvious.

PR #358, opened 2020-03-24
354 add individual photo dates

HilaryN:
I wasn't able to pull in the datetime as a string as implied in the api documentation - I had to make it a long in order to get it to work.
I've added a test for datetime in RetrofitApiClientTest.testGetPhotos and got the unit test to work by editing photos.json and adding in the datetime in there.  I added the correct datetime in long format for the photo which was being tested on (the last one in photos.json) and a dummy datetime (123) for the others.  Let me know if there is a different way to update the photos.json file rather than manually editing it as I have done.


HilaryN:
First image is on emulator (Pixel 2) (American date format); second is on my phone (UK format).

Emulator datetime test

Screenshot_FP2 datetime test 20200414-140805

oliverlockwood:
@HilaryN thanks for making those changes and for the screenshot.

No need to formally request a re-review, I get notified of pushes to the open PR and can re-review myself.

With the screenshots visible I personally think it would look better to set the `gravity` for the new field to be centered, to match the caption, or alternatively perhaps with it right-aligned.  What do you think?  I'm not going to mandate this change, but I'd like to know your thoughts.

oliverlockwood:
there's also now a merge commit you'll need to resolve, by pulling down `master` branch of the origin (not your fork).

HilaryN:
I'm happy to take suggestions on design as it isn't my strong point.
I've a feeling right-aligned will also look out of alignment with the caption, so centred may be best.  I'll take a look.

HilaryN:
After merging the changes from upstream master I am getting an error when trying to build:
"...AndroidStudioProjects\android\cyclestreets.app\license.properties (The system cannot find the file specified)"

oliverlockwood:
@HilaryN run `./ciLicense.sh` in the root.

If you're on Windows, then instead just manually perform the platform-equivalent copy operations in the `else` branch of that file.

Apologies for the inconvenience.

HilaryN:
Centre definitely looks better.  These are examples of centre and right-alignment on Pixel 2 API 28 emulator.

Screenshot_1587042407
Screenshot_1587042418
Screenshot_1587042506
Screenshot_1587042810
Screenshot_1587042784
Screenshot_1587042830

HilaryN:
This is on Nexus One API 29 emulator.
Screenshot_1587043328

HilaryN:
On my phone, photo picked at random... OK, I've heard of Boris bikes, but not this... @mvl22, nomination for photo of the month?

Screenshot_20200416-152543

mvl22:
Yes, it's quite fun to go in actually. It's informally known as the Boris Dangleway :)

Thanks for adding this feature. Slight improvements:

- Date could have the ordinal added, e.g. 13th
- You might want to format the month with its full name, e.g. September
- Lastly, ideally move it up slightly - it's a general principle of graphic design to avoid UI elements being right next to the edge of a line. (The same problem actually exists between the title and the photo.)

HilaryN:
I've never been on it, but the nickname made me laugh..  l'll add it to my list.  Can you take your bike on it?  ;-)

Regarding the ordinal on the date, I can't see an elegant way of doing it (i.e. an existing method) which wouldn't complicate translation into other languages and add to the maintenance burden?

The other two changes I'm happy to make.  (Will the last one be superseded by the new design?)

si-the-pie:
Yes, bikes can be taken on the Airline, here's a bit of an old link:

http://realcycling.blogspot.com/2012/07/thames-crossings-27a-emirates-air-line.html

HilaryN:
To think that I worked in Greenwich town centre (and even cycled up to N Greenwich one lunchtime) and never went on it :-)  I did at least go through the foot tunnel.

Here is the long month and extra spacing, subject to approval of code:

Screenshot_1587139396

oliverlockwood:
Looks good @HilaryN 👍 

I've pushed one additional commit, to credit you as a contributor.  This is now approved, I'll merge it once the Travis build finishes.

Note that the Google Play console currently says:
>Due to adjusted work schedules at this time, we are currently experiencing longer than usual review times
>Please expect review times of seven days or longer.

so it may not immediately appear!

mvl22:
> Here is the long month and extra spacing

Nice!

oliverlockwood:
Actually, this is already published, so if you're on the beta programme (@HilaryN you might want to opt in if you haven't already) then you can download the update and see it working, happy days!

PR #360, opened 2020-04-14
Update Gradle and various key plugins

### LINKED ISSUE #359: Update to latest Google APIs for artifact publication
Currently, master builds fail in the `ciPublish` stage with the following error:

```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':cyclestreets.app:publishApkRelease'.
> com.google.api.client.googleapis.json.GoogleJsonResponseException: 410 Gone
  {
    "code" : 410,
    "errors" : [ {
      "domain" : "androidpublisher",
      "message" : "This version of the Play Developer API is no longer available. For more information see https://android-developers.googleblog.com/2019/03/changes-to-google-play-developer-api.html.",
      "reason" : "apiDeprecated"
    } ],
    "message" : "This version of the Play Developer API is no longer available. For more information see https://android-developers.googleblog.com/2019/03/changes-to-google-play-developer-api.html."
  }
```

Looks like I need to read https://android-developers.googleblog.com/2019/03/changes-to-google-play-developer-api.html and fix this :)

oliverlockwood:
This should close #359 (by updating the version of the play-publisher plugin, which should itself target the newer Google APIs.)

I also took the opportunity to simplify some of our configuration files etc, as follows:
- config of key SDK versions etc is now in the root `build.gradle` file, so we don't have to load it in separately
- stuff that's specific to an application is now in the application's build.gradle file rather than being in the root (to make scoping easier to understand - it's not as if we're publishing multiple apps these days)
- reference to `cyclestreets.properties` to do manifest overrides has been removed - since I started managing the builds and publishing through Travis, we've never had such a file, so I presume it's not really necessary.

@jezhiggins I'd be grateful for your review when you have the chance, please.


cyclestreets.app/license.properties
```diff
@@ -0,0 +1,4 @@
+key.store=keystore
```
oliverlockwood, line 1: the attached keystore is valid, but contains a single dummy key

oliverlockwood:
@HilaryN feel free to review this and offer your thoughts.  This is why I recommended against upgrading gradle & other plugins etc. for your first activity 😉

mvl22:
Thanks for this, Oliver!

Just to be clear, this shows that we are targetting Android 6.0 as minimum level, yes?

oliverlockwood:
@mvl22 yes; that in itself is not a change, I've just refactored where we define the relevant settings.

oliverlockwood:
I'm going to merge this ahead of it being reviewed by @jezhiggins, so I can verify the publishing works as it should.  I will retrospectively address any feedback subsequently received, so feel free to comment with your thoughts.

HilaryN:
@oliverlockwood since then every time the little box pops up suggesting I upgrade something in Android Studio, I quiver in fear... :-D

PR #365, opened 2020-04-20
363: Fix waypoint icon rendering

### LINKED ISSUE #363: Waypointing has become a bit of an "art"... and the icons have become tiny
This is a screenshot from my phone (Pixel 2).

I did my damndest to put a waypoint on the junction between Church Street and High Street / The Broadway.  It didn't go so well location wise, and the icon is smaller than it used to be.

issue-waypointing

I suspect that the size, at least, is likely because the wisps are in the density-independent drawable folder (something which `lint` has been complaining about for ages) and I have a higher resolution on this phone than the previous.  So a starting point will be to fix that, and see if it makes a difference to the locations or not.

@mvl22 @jezhiggins where did the wisps originally come from?   (I'm talking about e.g. https://github.com/cyclestreets/android/blob/master/libraries/cyclestreets-view/src/main/res/drawable/green_wisp.png)
If we have some higher-resolution versions of these and/or scalable vector representations, that would be helpful.

oliverlockwood:
Closes #363.

Notes:
1. ~~I've moved all the SVG files representing the waypoint wisps, that @si-the-pie committed, to the root `assets` folder.  That way, they don't get built into the APK (and in any case, SVG files are not considered valid in themselves; their presence was breaking the build.)~~ This is now already done on `master` branch.
2. We need the point of the wisp to be centred on the image we use in the app.  Although this can be changed for an SVG by editing the `viewBox`, and similarly for imported Android drawable XMLs, for some reason when rendering vector wisps directly in the app the border stroke is omitted (see screenshot below; ignore scaling; the orange wisps are the ones directly using vectors).  As a result I've imported into PNGs at sensible scale, and direct use of vectors can wait for another day.
   Screen Shot 2020-04-20 at 07 58 20
3.  I've extended our code to handle rendering vector drawables, even if we're not using them right now.
4.  I have confirmed that the scaling issue was the cause of the off-location waypoint drawing, also.  So this PR should resolve those aspects.
5.  As ever, I've done a little Java -> Kotlin migration in place.


oliverlockwood:
@jezhiggins would you cast your eye over this when you have the chance, please?

oliverlockwood:
Merging so I can move forward with fixing some other things.  @jezhiggins feel free to add comments when you get the chance to look; I will be notified of these and I don't mind dealing with them retrospectively.

PR #367, opened 2020-04-25
364: Fix back-into-settings weird overlay bug

### LINKED ISSUE #364: Pressing "back" into the Settings window causes weird transparent overlay
See screenshot, taken on my phone:

issue-settings-overlay

Also reported by a user at https://play.google.com/apps/publish?account=7698300952134656507#ReviewDetailsPlace:p=net.cyclestreets&reviewid=gp:AOqpTOGjBMNC2vKaxCS0fmlRAt9YMB39sajnCHCw84fHH1NmBWJhNBL2RwCXNgiZiT8ltq5BXA3630KkjXx4uw

I believe this came in with the major menu updates that I did in 3.6.  I remember seeing it occasionally but I ran out of time to investigate.
We should fix in 3.7.

oliverlockwood:
Closes #364.

We still want to be able to use "back" while _within_ the Settings menu.  The approach I've taken is: when the user navigates out of the Settings menu (which, to my knowledge, can only be done via the main nav), any back-stack entries relating to the Settings menu are dropped.

I've tested fairly thoroughly on the emulator and it seems to do the right thing.

@jezhiggins, would you please review?

### DIFF
--- a/cyclestreets.app/src/main/assets/whatsnew.html
+++ b/cyclestreets.app/src/main/assets/whatsnew.html
@@ -21,6 +21,7 @@
   <ul>
     <li>LiveRide no longer restarts on wake, so the annoying torrent of instructions is gone.</li>
     <li>Waypoint icons in Android 10 no longer appear ridiculously small and in the wrong place.</li>
+    <li>Navigating back after having previously used the Settings screen no longer results in a weird transparent menu overlay.</li>
     <li>Extra start waypoint added when replanning the journey does not persist, so multiple replans are dealt with more logically.</li>
     <li>A number of crash scenarios fixed.</li>
     <li>LiveRide now makes itself a foreground service, which fixes some scenarios where it previously wouldn't work (e.g. low available memory)</li>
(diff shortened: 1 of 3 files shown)


jezhiggins:
I've not seen this problem myself, which makes me wonder if it's confined to a particular combination or combinations of Android version, screen size/density, etc.

That said, I've no problem with the code, but given https://github.com/cyclestreets/android/issues/368#issuecomment-619543159 they may not be necessary so please hang fire for a moment.

oliverlockwood:
Sounds good, happy to hold fire.

For information, the issue reproduces on the latest (beta) version of the codebase on a Pixel 2 emulator running Android 10. It definitely is evident on other versions too, just following the instructions on the linked issue.

oliverlockwood:
Also FYI I'm pretty sure it was introduced when I modernised the Settings menu to simplify, taking advantage of native Android functionality that wasn't there when you first wrote it :-)

oliverlockwood:
Superceded by #369.

PR #371, opened 2020-04-29
295: Point Blog URL to cyclestreets.org/news

### LINKED ISSUE #295: Blog formatting
As spotted by @DoccyB (CycleStreets's summer intern 2018) the title text overflows:

screenshot_20180829-153406_cyclestreets

jezhiggins:
Prevents the redirect, so we stay with the app.

Fixes #295, specifically https://github.com/cyclestreets/android/issues/295#issuecomment-616092298


jezhiggins:
Screenshot_1588198416

jezhiggins:
Ideally that header image would be a little smaller - even on a big tablet it's pretty dominating - but that's probably best dealt with on the server-side with a tweak to the bottom-header CSS - either generally, or device specific - @mvl22 can you pick that up?

mvl22:
Yes, please don't deal with this client-side.

mvl22:
Is there any checking of the feed, by the way?

If so that RSS link should also be updated.

PR #374, opened 2020-05-01
Upgrade various dependencies

### LINKED ISSUE #373: Bump osmdroid
I see that osmdroid is now at 6.1.6:
https://github.com/osmdroid/osmdroid/releases

Might be worth bumping up further, as that will presumably pull in 18 months of bugfixes and performance improvements.

oliverlockwood:
Closes #373.

(Work in progress - I need to check this actually compiles and runs...)


oliverlockwood:
I've run this in the emulator and it seems to generally work, so I think we can probably go with it.

Issue #375, opened 2020-05-01
Back button behaviour requires lots of clicks to minimise app when you have planned a journey


oliverlockwood:
Introduced by #369.

To reproduce:
* Go to Journey Planner
* Mark out a route with _n_ waypoints (e.g. 3)
* Click "Tap here to Route" to plan the journey
* Now use the Back button.  To minimise the app, you have to click Back _n + 1_ times (e.g. 4 in our example).

Although we're note removing waypoints because we're not in planning mode, I presume that some of the logic around click-back-to-remove-waypoints is still being processed, due to the _n + 1_ behaviour.

oliverlockwood:
Closed by #376

PR #376, opened 2020-05-01
Fix Journey Planner Back button behaviour

jezhiggins:
TapToRoute was rebuilding the undo stack on every restore, but should only do so when journey planning is still in progress. Fixed. 
Partly planned journeys could be lost if the app was paused and resumed. Also fixed.

Addressed issue #375


libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    private fun setRouteLoaded() {
+        prefs()
+                .edit()
+                .putBoolean(routeLoadedPref, true)
+                .remove(waypointsInProgressPref)
+                .apply()
+    }
+    private fun clearRouteLoaded() {
```
oliverlockwood, line 180: I find code generally easier to read with a blank line between consecutive methods.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    private fun stashWaypoints() {
+        val stash = RouteDatabase.serializeWaypoints(waypoints_)
+        prefs()
+                .edit()
+                .putString(waypointsInProgressPref, stash)
+                .apply()
+    }
+    private fun restoreWaypoints() {
```
oliverlockwood, line 194: as above - blank line between methods?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    interface Listener {
+        fun onNewJourney(journey: Journey, waypoints: Waypoints)
+        fun onResetJourney()
+    }
+
+    private class Listeners {
+        private val listeners_: MutableList<Listener> = ArrayList()
+        fun register(listener: Listener) {
```
oliverlockwood, line 217: I'd probably put a blank line before the method

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -0,0 +1,244 @@
...
+            if (!doRegister(listener)) return
+            if (journey() != NULL_JOURNEY || waypoints() != NULL_WAYPOINTS) listener.onNewJourney(journey(), waypoints()) else listener.onResetJourney()
```
oliverlockwood, line 219: can we please break this line up like:
```
if (test || test)
  doThing()
else
  doOtherThing()
```

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -0,0 +1,244 @@
...
+            listeners_.add(listener)
+            return true
+        }
+
+        fun unregister(listener: Listener) {
+            listeners_.remove(listener)
+        }
+
+        fun onNewJourney(journey: Journey, waypoints: Waypoints) {
+            for (l in listeners_) l.onNewJourney(journey, waypoints)
```
oliverlockwood, line 237: Could this be  like`listeners_.forEach(...)`?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        fun onReset() {
+            for (l in listeners_) l.onResetJourney()
```
oliverlockwood, line 241: listeners_.forEach(...)?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
jezhiggins: I had both these as single line methods, then jiggled them about and wasn't paying proper attention.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
jezhiggins: Certainly can

oliverlockwood:
Looks good to me, and hacking about on the emulator I haven't been able to prove something is massively broken 😄

PR #378, opened 2020-05-01
Audit cyclestreets.net/cyclestreets.org links

### LINKED ISSUE #372: Ensure HTTPS resources referenced rather than HTTP
Over the last few years, we've moved all our infrastructure over to HTTPS.

So things like the blog, cycle.st redirects, the API, etc., are all on HTTPS.

It would be good to ensure that resources are directly using the HTTPS URLs, to avoid redirects slowing requests down, and that links to other sites, e.g. openstreetmap, are similarly updated (as they are mostly on HTTPS too).

There are still some references to HTTP, to work through:
https://github.com/cyclestreets/android/search?q=http+-filename%3A*.svg

(Obviously that search picks up some things such as internal Android code libraries which mention http, and those should obviously not be touched.)

jezhiggins:
Make sure they're all still valid, and using https.

Fixes #372


oliverlockwood:
I've got Java 8 here and can look into the UT failures, rather than you have to have a huge cycle time.  Stand by.

jezhiggins:
Thanks chum. I did look at moving to the latest version to see if that would resolve it, but that has a transient dependency on Jetpack and so we can't use it yet ....

oliverlockwood:
@jezhiggins I've fixed the test and corrected the cache-control override implementation.  All hail the unit tests, as once configured correctly they identified a miss where we would have lost the caching of the blog feed! 💪

jezhiggins:
Good stuff. That's brilliant. Thank you 🙂On 1 May 2020 19:53, Oliver Lockwood <notifications@github.com> wrote:

@jezhiggins I've fixed the test and corrected the cache-control override implementation.  All hail the unit tests, as once configured correctly they identified a miss where we would have lost the caching of the blog feed! 💪

PR #381, opened 2020-05-03
Fix NumberFormatException in Route.restoreWaypoints

jezhiggins:
It's possible that the stashed waypoints string might be empty, rather
that null. In this case, attempting to restore waypoints at app start
will throw a number format exception and the app will crash.

This patch corrects that.

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
@@ -195,7 +195,7 @@ object Route {
 
     private fun restoreWaypoints() {
         val stash = prefs().getString(waypointsInProgressPref, "")
-        if (stash.isNotEmpty()) {
+        if (stash != null && stash.isNotEmpty()) {
             waypoints_ = Waypoints(RouteDatabase.deserializeWaypoints(stash))
         }
     }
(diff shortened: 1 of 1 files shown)


jezhiggins:
@oliverlockwood This is entirey on me and I'm sorry I didn't pick it up before. I only realised what was going on when the new build came through on one of my devices. If this patch is ok with you can you merge and release as soon as you can? Thanks.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -194,9 +194,9 @@ object Route {
     }
 
     private fun restoreWaypoints() {
-        val stash = prefs().getString(waypointsInProgressPref, null)
-        stash?.let {
-            waypoints_ = Waypoints(RouteDatabase.deserializeWaypoints(it))
+        val stash = prefs().getString(waypointsInProgressPref, "")
+        if (stash.isNotEmpty()) {
```
oliverlockwood, line 198: If stash *is* null - e.g. because someone's run the previous version, where it would default to null - won't we now have an NPE?

I suggest checking both in this `if` test for safety.

oliverlockwood:
@jezhiggins one inline comment - I think you should change this to avoid risking an NPE.

Once you've made the change, I'll happily issue another release - I'll call it 3.7.1, to avoid ambiguity - and I'll be sure to update the F-droid definition accordingly too.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
jezhiggins: I don't believe it can be null - it might be the empty string, or it might not be set at all (in which case we now default to the empty string) - but I buggered it up last time, so yes, you're right.

PR #382, opened 2020-05-05
Localisation updates

andrewshadura:
Hi,

This pull request extracts more strings into translatable resources and fixes punctuation in some source strings and corresponding translation.

While at it, it would be great if you uploaded these translations to Transifex, since it’s missing some of them (e.g. Russian).

Thanks!


mvl22:
Nice, thanks!

andrewshadura:
Speaking of which, the Russian translation needs an overhaul since it reads quite unnatural. I will submit updates as soon as it is on Transifex since it’s a bit inconvenient to do that across multiple XML files.

oliverlockwood:
Hi @andrewshadura - thanks for submitting this.  We'll get to reviewing this in due course when we've dealt with some other issues.

andrewshadura:
@oliverlockwood, could you please also push all translations to Transifex?

Issue #316, opened 2018-09-03
Migrate support library use to AndroidX


oliverlockwood:
Not right now... but when it becomes GA (it's in RC now)

https://android-developers.googleblog.com/2018/05/hello-world-androidx.html

jezhiggins:
Planning to look at this shortly. There's tool support for the change, and AppCompat stuff is EOL.

oliverlockwood:
This was closed by #383.

PR #383, opened 2020-05-05
Migrate to jetpack

jezhiggins:
Migrating to the androidx packages, along with other dependencies we've had to hold back on.

Don't propose to merge this in while we're still fixing bugs in the current release. 

Addresses issue #316


libraries/cyclestreets-core/build.gradle
```diff
@@ -9,8 +9,6 @@ dependencies {
   api 'org.osmdroid:osmdroid-android:6.1.6'
   api 'org.slf4j:slf4j-android:1.7.30'
 
-  api "com.android.support:support-compat:${rootProject.ext.supportLibraryVersion}"
```
oliverlockwood, line 12: the definition of `supportLibraryVersion` can probably be removed from the root `build.gradle`

libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/LocationsFragment.java
```diff
@@ -3,8 +3,8 @@
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
-import android.support.design.widget.FloatingActionButton;
-import android.support.v4.app.ListFragment;
+import com.google.android.material.floatingactionbutton.FloatingActionButton;
+import androidx.fragment.app.ListFragment;
```
oliverlockwood, line 7: Might be worth auto-reordering imports for clarity in this and all the classes you've changed.

libraries/cyclestreets-view/build.gradle
```diff
@@ -18,13 +18,9 @@ dependencies {
...
-  api "com.android.support:appcompat-v7:${rootProject.ext.supportLibraryVersion}"
-  api "com.android.support:design:${rootProject.ext.supportLibraryVersion}"
-  api "com.android.support:exifinterface:${rootProject.ext.supportLibraryVersion}"
-  api "com.android.support:preference-v7:${rootProject.ext.supportLibraryVersion}"
-  api "com.android.support:preference-v14:${rootProject.ext.supportLibraryVersion}"
-  api "com.android.support:support-annotations:${rootProject.ext.supportLibraryVersion}"
-  api "com.android.support:support-v4:${rootProject.ext.supportLibraryVersion}"
+  api 'com.google.android.material:material:1.0.0'
+  api 'androidx.exifinterface:exifinterface:1.0.0'
+  api 'androidx.preference:preference:1.0.0'
```
oliverlockwood, line 23: Is it worth extracting the Android X version (seemingly `1.0.0`) to a constant in the root build-gradle?

oliverlockwood:
@jezhiggins I've made a bunch more commits here, on top of your changes, and have got as far as a green build and an application which starts up.

From a brief skim through in the emulator, most things seem to work correctly; one aspect that's weird is the menu items on the Settings screen seem to be arbitrarily shifted to the right.

However, from looking at https://source.android.com/devices/tech/settings/settings-guidelines, this seems to be the new convention, so maybe it's OK.

What more do you think needs doing on this before we merge it, and send it out to our beta user base?

jezhiggins:
In my mind, I had the idea that we could roll things like you've been doing in #396 into this PR as well. Updating the robolectrics dependency, like the iconics library, also depends on the move to JetPack. Maybe it is best done in two steps - push this out, then if there are problems we can resolve/unwind those separately from anything that might arise from the third party dependencies.

mvl22:
> However, from looking at https://source.android.com/devices/tech/settings/settings-guidelines, this seems to be the new convention, so maybe it's OK.

I see that the menu items are now aligned to the title, which is more like what a designer would probably do. I assume the intention there will be icons.

I wouldn't worry too much about this, as we obviously plan to get the full design stuff in in due course. (The implementation of this on the mobile website is moving at pace, by the way.)

oliverlockwood:
@jezhiggins yep, I'd rather split the iconics and robolectric upgrades into another PR.

So I'll give this an approval - are you happy for this to go in?

jezhiggins:
Yep :+1:

PR #389, opened 2020-05-06
Fix pinch-zoom in LiveRide

### LINKED ISSUE #384: Pinch zoom in/out no longer works while in LiveRide
As reported by a beta user:
>The current beta version seems okay, except that I can't zoom in or out during LiveRide, which would be useful if I am coming up to a complicated junction.

I experienced this myself yesterday (on my Pixel 2, running Android 10, during my daily allocation of exercise), and it's definitely a regression.  May have come in when I upgraded `osmdroid`; I'm not sure.  This should be fairly easy to reproduce on an emulator, and thence fix.

Interestingly you can still zoom _in_ by double tapping, but zooming in or out via pinching is not possible.

jezhiggins:
When locked on location, ignore single-finger pan events. Allow
multitouch through, so the gesture detector can pick up pinch-zoom
properly.

You do get a little bit of panning during a pinch-zoom (and also
double-tap zoom), but the next location update will recentre the map.

Fixes #384.

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.kt
@@ -30,7 +30,6 @@ class LocationOverlay(private val mapView: CycleMapView) :
     private val button: FloatingActionButton
     private val onColor: Int = Theme.lowlightColor(mapView.context)
     private val followColor: Int = Theme.highlightColor(mapView.context) or -0x1000000
-    private var lockedOn: Boolean = false
 
     private class UseEverythingLocationProvider(context: Context) : GpsMyLocationProvider(context) {
         init {
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.kt
@@ -90,20 +90,23 @@ class LocationOverlay(private val mapView: CycleMapView) :
     }
 
     fun lockOnLocation() {
-        lockedOn = true
+        setEnableAutoStop(false)
     }
 
     fun hideButton() {
         button.visibility = View.INVISIBLE
     }
-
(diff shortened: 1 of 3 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.kt
```diff
@@ -90,20 +90,23 @@ class LocationOverlay(private val mapView: CycleMapView) :
     }
 
     fun lockOnLocation() {
-        lockedOn = true
```
oliverlockwood, line 93: as far as I can tell, `lockedOn` is now an unused variable, can we remove it please?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.kt
```diff
@@ -90,20 +90,23 @@ class LocationOverlay(private val mapView: CycleMapView) :
...
+        setEnableAutoStop(false)
     }
 
     fun hideButton() {
         button.visibility = View.INVISIBLE
     }
-
     override fun onTouchEvent(event: MotionEvent, mapView: MapView?): Boolean {
```
oliverlockwood, line 99: Is it worth adding a comment to this method explaining what we're doing and why, perhaps linking to our Github issue and/or (if you raise one) the issue on `osmdroid`?  Potentially then we could rip out code if/when you or someone else fixes this in `osmdroid`.

jezhiggins: I'll add a link back to this issue. If the change does go into osmdroid, we would be able to remove this onTouchEvent entirely.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LocationOverlay.kt
jezhiggins: Yes, you're right. It's exactly this kind of mistake I was trying to avoid by getting some sleep ... Should have made a note.

Issue #391, opened 2020-05-07
(Optionally) rotate the map with travel direction in LiveRide


oliverlockwood:
We occasionally get feedback about this, e.g. the following excerpt from a review:
>Would be better if the map rotated with travel direction

A brief search on the internet threw up this page: https://stackoverflow.com/questions/37575522/osmdroid-how-to-rotate-osm-map-as-per-devices-direction which suggests this isn't too infeasible to achieve; bearing in mind we're currently on the latest version of osmdroid it's probably worth having a go.

jezhiggins:
OsmDroid does now support this natively - previously, IIRC, you had to do the maths (not a huge problem) and the translation (pita) yourself. Curiously the only OsmDroid sample that demonstrated the pinch-zoom issue, https://github.com/osmdroid/osmdroid/issues/1578, also does map rotation, so a quick squint at https://github.com/osmdroid/osmdroid/blob/master/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/location/SampleHeadingCompassUp.java should get us going.

jezhiggins:
Call is mapView.setMapOrientation(degrees)

There's a tiny bit of adjustment to do to correct for the device portrait/landscape orientation. You can use the bearing from the location provider or from the compass.

jezhiggins:
Oh - MapView now has setMapCenterOffset(x, y) which I think means we could move the marker down the screen, so as we rotated the you see more of the route in front of you.

mvl22:
> OsmDroid does now support this natively

Awesome. This would be a great improvement to add.

jezhiggins:
Very rough and ready first go, showing rotation with offset center. This has had no testing other than me walking up and down my garden, but I'll go for a walk round the block with it tomorrow to see how it holds up. 

![Screenshot_20200508-224644](https://user-images.githubusercontent.com/555337/81452456-df828080-917e-11ea-97bf-357ecf5863f4.jpg)

For reference - Anderton Park Road runs NE/SW. You can see that the wisp has also be rotated. We'd probably need to apply a reverse translation drawing those.

mvl22:
Does the vector map also rotate, and fix the label alignment? Obviously a bitmap background can't do that, as shown.

jezhiggins:
The rotation is done by the map view as it paints the screen, so is independent of the tile source. Consequently, the answers to your questions are yes, but no.

jezhiggins:
[...]

PR #392, opened 2020-05-08
Rotate map in liveride

jezhiggins:
Dashed off overlay to support map rotation, based on OsmDroid sample code. Appears ok on bench test, but can't vouch for it beyond that.

Addresses #391


libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/RotateMapOverlay.kt
```diff
@@ -0,0 +1,174 @@
...
+        rotateButton = rotateButtonView.findViewById(R.id.compass_button)
+        rotateButton.setOnClickListener { setRotation(!rotate) }
+
+        mapView.addView(rotateButtonView)
+        locationProvider = UseEverythingLocationProvider(context)
+        compassProvider = InternalCompassOrientationProvider(context)
+    }
+
+    private fun setRotation(state: Boolean) {
+        Log.d("LiveRide", "Setting map rotation to $state")
```
oliverlockwood, line 63: Personally I prefer to put a static class `TAG` to avoid potentially repeating log strings like this.

jezhiggins:
The UseEverythingLocationProvider is common with the LocationOverlay. Last night, I couldn't think of a suitable package to put that in, but I'll think of something.

oliverlockwood:
I'm going to merge this and get it onto the beta so I can test it on my phone easily (and so can the rest of the beta world).

Issue #393, opened 2020-05-09
LiveRide speech tweaks


jezhiggins:
Should announce metres and kilometres in full. Currently announces "Continue 100m" as "continue one hundred em" which is a bit rubbish.

If possible, I'd like to try and get road designations announced properly. "Wheeler's Lane A4040" is announced "Wheeler's Lane ay four thousand and forty" and "Castle Hill Road B1113" is announced "Castle Hill Road bee one thousand and thirteen." In an ideal world, we'd be able to transform those to "ay forty forty" and "bee treble one three", but even "ay four oh four oh" and "bee one one one three" would sound significantly more natural.

oliverlockwood:
I have thought exactly the same thing.

oliverlockwood:
I started looking at a regex to solve this but realised that our parlance is a little more complicated than that... so before I implement it, here's some proposed logic around this:

**1 or 2 digits** - leave & pronounce normally

**3 digits**
- if ends with 00, leave as-is, e.g. `A400` => pronounced as "Eh Four Hundred"
- else if ends with 0, split into single digit then double, e.g. `A230` => `A2 30` => pronounced as "Eh Two Thirty"
- otherwise, split into single digits, e.g. `B467` => `B4 6 7` => pronounced "Bee Four Six Seven"

**4 digits**
- if ends with 000, leave as-is, e.g. `A3000` => pronounced as "Eh Three Thousand"
- else if ends with 00, split off the "hundred", e.g. `B1200` => `B12 hundred` => pronounced as "Bee Twelve Hundred"
- else if either or both two digit separation ends with a zero, tokenise into two two-digits numbers,
   - e.g. `A1450` => `A14 50` => pronounced as "Eh Fourteen Fifty"
   - e.g. `B2031` => `B20 31` => pronounced as "Bee Twenty Thirty-One"
- else if any triple digits occur, sort those out, e.g. `B1113` => `B Triple 1 3`=> pronounced "Bee Triple One Three"
- otherwise, split into single digits, e.g. `A4123` => `A4 1 2 3` => pronounced as "Eh Four One Two Three"

@jezhiggins what do you think to that approach?

I should be able to keep this nicely isolated (and therefore unit-testable) by extending and extracting the `LiveRideState.speechify()` method.

oliverlockwood:
PR now available for the metres / kilometres bit; I'll work on the road number implementation as an extension in due course.

jf1:
[...]

PR #394, opened 2020-05-10
393: fix pronounciation of metres and kilometres

oliverlockwood:
Towards issue #393.

(doesn't close it, because there's still some open discussion about how to handle road numbers).


jezhiggins:
This is such a pleasing change :)

PR #400, opened 2020-05-12
399: Fix LiveRide

### LINKED ISSUE #399: LiveRide crashes the app (after upgrading AndroidX/Jetpack)
'Nuff said...

I'm looking at this one now

oliverlockwood:
Fixes #399.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-  ext.kotlinVersion = '1.3.61'
+  ext.kotlinVersion = '1.3.72'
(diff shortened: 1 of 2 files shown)


oliverlockwood:
I've tested this locally and, through painful manual setting of the location, have managed to prove that LiveRide _functionally_ seems to work, as well as just starting up correctly.

I'm going to self-merge this so we don't leave our beta users with a broken build.  As ever @jezhiggins feel free to make comments which I can retrospectively address.

Issue #402, opened 2020-05-14
Update icon set for POIs and Photomap


oliverlockwood:
See conversation on #312 for context.

oliverlockwood:
POI icons have now been upgraded by #403.  These are actually derived from what's returned to us over the REST API; in this PR I changed our code to request 64x64 icons instead of 32x32, but so as to still display them at the same size on the screen (thereby we have just improved the resolution of these).
 
No work was done to improve the Photomap icons, and these still look a bit blurry.  I'll see what we can do about this in due course.

mvl22:
Just to flag up that the Photomap API also returns icons, as SVG, and ideally you would therefore just use the `iconUrl` field for those, caching the path, rather than using hard-coded ones in the app distribution.

oliverlockwood:
@mvl22 thanks.  I'm not sure whether Android will cope with downloading and then rendering an SVG though.  If you import a vector asset into your drawable folder Android will convert it from SVG to its own vector-representation XML (which is what we've done with the direction arrows etc.)

So we may end up with a choice between:
- embedding vector images in the distribution
- downloading and using PNGs

mvl22:
Please embed then.

mvl22:
The icons appear very large on a Motorola G6 Play:

![Screenshot_20200519-144629](https://user-images.githubusercontent.com/361423/82843260-892e7500-9ed4-11ea-837e-b056f51d3b04.png)

oliverlockwood:
hmm, I'll try specifically getting an emulator of that device, and will see what's going on.

oliverlockwood:
@mvl22 what version of Android is running on this?  perhaps you can screenshot the phone's info screen or similar?

oliverlockwood:
While I'm waiting for feedback on the above issue, I've been experimenting with the vector graphics for Photomap icons.
Here's a couple of screenshots.  In both, I'm only using vector graphics for the "cycleways - neutral" image.

In the first, I've worked to get the icons as close to the same size as possible (by setting the drawable height to 17dp).

In the second, I've offered an alternative size (height 21dp - a nicely scaled half; the proportions of the original vector definition are 38w x 42h).

<img width="433" alt="Screen Shot 2020-05-29 at 20 36 29" src="https://user-images.githubusercontent.com/6017680/83298442-5e9e2e00-a1ec-11ea-91fb-0a202c6adbd1.png">
[...]

PR #403, opened 2020-05-17
The Great Icons Upgrade

### LINKED ISSUE #366: Use new vector images for directions
Follows on from #363, alongside with @si-the-pie was kind enough to upload a bunch of vector assets for our direction arrows etc.  These can be found in the `/assets/directions` folder of the repo.

This ticket covers making use of those in the codebase and removing the old density-independent bitmaps.

oliverlockwood:
Now ready for review.  Comprising the following key areas:
### Iconics:
*Upgrading the Iconics libraries to latest
* Moving away from deprecated usage of IconicsContextWrapper (see mikepenz/Android-Iconics#470)
* Replacing some more blurry old icons with Material design ones from the Iconics library
* Refactoring the `IconicsHelper` methods for more concise usage

### And the rest:
* Replacing all the navigation icons used in LiveRide with @si-the-pie's new vector ones (closes #366)
* Adding versions of the compass icons for use in different pixel density screens
* Moving a bunch of SVG images which are unused, but may be useful for reference, out into the root `assets` folder, so as not to be part of the packaged app
* Removing a whole lot of unused images
* Changing a couple of classes from Java to Kotlin
* Removing the unused `osmarender` / `rendertheme` stuff - I believe this is a hangover from when we used to package the Mapsforge library instead of including it as a dependency

### Added since PR was originally raised:
* Making POI icons less blurry (towards #402)
* Bug identified and fixed: POI category list view had gone a bit mental when re-entered after previously selecting something
* Icons added to the root Settings screen (see screenshot at the bottom)

## Descriptive screenshots
The only thing that can be considered a _change_ instead of purely an _improvement_ is my decision to replace the lock-screen-on button, from just being an open padlock, to this:
Screen Shot 2020-05-17 at 16 35 51

New icons on the Settings screen are shown here:
Screen Shot 2020-05-18 at 16 36 24


mvl22:
Sounds great!

I'll look forward to testing this tomorrow (I have a new second-hand Android 9 device arriving, for testing purposes!)

mvl22:
Does this automatically result in a new build in Google Play appearing on the phone?

oliverlockwood:
Thanks @jezhiggins.

@mvl22 after being merged the artifact goes straight to Google Play, but in my experience it normally takes a few hours before it appears as an available update to a phone which is on the Beta programme.  Speaking personally, I merged this 4 hours ago, and have been pinging the Play store intermittently since, and it's only just appeared to me as available.  (But it may be made available in blocks to different ranges of user IDs; I don't know what Google does under the covers here.)

But it's worth you refreshing the Play Store now and taking a look.

mvl22:
OK, so just to be clear, this happens automatically:

Push to master -> Automated build -> Google Play beta store

oliverlockwood:
@mvl22 yes. I think we've been over this a few times now... 🙂

PR #410, opened 2020-05-24
Calculate and display remaining journey in LiveRide.  Shift attribution.

HilaryN:
Would you be able to have a look over this and give me feedback?  

Functionally it works, but I still need to test battery life as the other day when using it my battery went dead after less than 40 minutes (I would normally expect an hour in LiveRide).  I hope to test pre- and post- versions this week.  I know I could use tools but I'd like to do it for real :-)

Also I haven't done the unit tests yet.

Thanks!

[Test remaining journey.pdf](https://github.com/cyclestreets/android/files/4674073/Test.remaining.journey.pdf)


HilaryN:
The build failed on Travis because of not having the translations - is that right?

oliverlockwood:
@hilaryn I will have a look either tonight or tomorrow night.

HilaryN:
I did an hour of LiveRide pre- and post-changes and it used less than 50% battery in both cases so that seems OK.  I used a different battery this time (the original).  
The time when it went flat I had stopped at my destination (25mins) for over an hour, though only occasionally using my phone for stuff.

oliverlockwood:
@HilaryN yes, the build failed on Travis because of missing translations.  If you add a translatable string but don't know how to translate it, that's fine, just put the string in the other languages with the English representation and we can go from there.

Remember that you should be able to reproduce any Travis build failures locally.  `.travis.yml` defines our Travis CI build process (in combination with the unspecified default behaviour that Travis provides - which is well documented on their website).  Also you can always scroll up in the Travis output to find the last command that was run (in this case, `./gradlew build`) which should give you what you need to reproduce and fix locally.

(Assuming of course, you haven't done anything crazy like upgrading all the Gradle setup locally without telling anyone 😉 )

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -44,6 +46,42 @@ class Journey private constructor(wp: Waypoints? = null) {
...
+
+    fun remainingDistance(distanceUntilTurn: Int): Int {
+        val actSeg = activeSegment()
+
+        if (actSeg != null) {
+            return if (activeSegment > 0) (totalDistance() - actSeg.cumulativeDistance + distanceUntilTurn)
+                // segments[0] is a summary of the whole journey and its Cumulative distance is same as Total distance
+            else (totalDistance()) // Don't try and calculate until actual segment has been found
+        }
+        else { return Int.MAX_VALUE }
```
oliverlockwood, line 59: Why not return `totalDistance()` in this case?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -44,6 +46,42 @@ class Journey private constructor(wp: Waypoints? = null) {
...
+    }
+
+    fun remainingTime(distanceUntilTurn: Int): Int {
+        val actSeg = activeSegment()
+        val prevSeg = previousSegment()
+        val prevSegTime : Int
+        val timeToEndOfSeg : Int
+
+        if (prevSeg != null) {prevSegTime = prevSeg.time}
+        else {prevSegTime = 0}
```
oliverlockwood, line 69: this is not nice to read - it looks like a closure but it's just the body of an if statement.

Please refactor - look at code around.  We're not 100% consistent, but we don't have any code in that's this yucky 😂  I think this should be like:
```kotlin
prevSegTime = if (prevSeg != null) prevSeg.time else 0
```

In fact, if you do it this way, you can inline it with the declaration of `prevSegTime`.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -44,6 +46,42 @@ class Journey private constructor(wp: Waypoints? = null) {
...
+
+        if (actSeg != null) {
+            // Time to cover whole of active segment (Segment.time is cumulative time)
+            val segTime = actSeg.time - prevSegTime
+            if (actSeg.distance != 0) {
+                // This is an approximation of the time to the end of the active segment from current location
+                timeToEndOfSeg = ((segTime * distanceUntilTurn).toFloat() / actSeg.distance.toFloat()).toInt()
+            }
+            else {timeToEndOfSeg = 0}
```
oliverlockwood, line 78: In this case I think
```kotlin
else {
    timetoEndOfSeg = 0
}
```
would be appropriate for readability, consistent with the style in the `if` (which in turn I think is correct)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
oliverlockwood: In fact I would change this whole block to simplify, like so:

```kotlin
if (actSeg != null && activeSegment > 0) {
    // segments[0] is a summary of the whole journey and its Cumulative distance is same as Total distance
    return (totalDistance() - actSeg.cumulativeDistance + distanceUntilTurn)
}

return totalDistance() // Don't try and calculate until actual segment has been found
```

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -44,6 +46,42 @@ class Journey private constructor(wp: Waypoints? = null) {
...
+
+            return if (activeSegment > 0) (totalTime() - actSeg.time + timeToEndOfSeg)
+            else (totalTime()) // Don't try and calculate until actual segment has been found
+        }
+        else { return Int.MAX_VALUE }
```
oliverlockwood, line 83: as above:
- I don't think it's appropriate to return `Int.MAX_VALUE`
- I would simplify logic so you have a block like
```kotlin
if (actSeg != null && activeSegment > 0) {
    ... all the other logic ...
    ... ending with a definite return...
}

return totalTime()
```

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -57,6 +95,9 @@ class Journey private constructor(wp: Waypoints? = null) {
 
     fun activeSegmentIndex(): Int { return activeSegment }
 
+    fun previousSegment(): Segment? {
+        return if (activeSegment > 1) segments[activeSegment-1] else null
```
oliverlockwood, line 99: There should be spaces around the `-` minus sign.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -134,6 +175,7 @@ class Journey private constructor(wp: Waypoints? = null) {
                 totalTime += sdo.time
                 totalDistance += sdo.distance
                 journey.segments.add(
+                        // Format time for display in Itinerary
```
oliverlockwood, line 178: Indentation of this Comment is odd.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Segment.java
```diff
@@ -53,6 +55,7 @@
           final String turnInstruction,
           final boolean walk,
           final String runningTime,
+          final int time,
```
oliverlockwood, line 58: Given the usage of this is in parallel with `cumulativeDistance`, should we not call it `cumulativeTime`?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
```diff
@@ -25,6 +25,7 @@
 public class TileSource {
 
   private static final String TAG = Logging.getTag(TileSource.class);
+  private static int boxHeight;
```
oliverlockwood, line 28: Why is the box - conceptually unique to the `LiveRideOverlay` - associated with the `TileSource`? That doesn't seem to make logical architectural sense.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LiveRideOverlay.java
```diff
@@ -60,7 +80,35 @@ public LiveRideOverlay(final Activity context, final Locator locator) {
...
+      // After 1st seg, text will be removed from bottom box and height reduced
+      heightNoText = height - smallLineHeight_;
+    }
+    TileSource.setBoxHeight(height);  // Use this height to shift map attribution in ControllerOverlay
+    if (showTime) {
+      titleVerticalPosition = offset_ + midLineHeight_ * 2;
+    }
+    else {
+      titleVerticalPosition = (largeLineHeight_ - smallLineHeight_ - midLineHeight_ - offset_) / 2 + midLineHeight_ + offset_;
+    }
```
oliverlockwood, line 111: This new block of code is too much to put into the constructor.  Please contextualise.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LiveRideOverlay.java
```diff
@@ -172,15 +238,123 @@ private String speed() {
     return formatter_.speed(location.getSpeed());
   }
 
-  private String distanceUntilTurn() {
+  private int distanceUntilTurn() {
+    final GeoPoint whereIam = whereIAm();
+    final int fromEnd;
+    if (whereIam == null) {
+       fromEnd = Integer.MAX_VALUE;}
```
oliverlockwood, line 245: Three things here:
1. Why are you using `Integer.MAX_VALUE` as a magic number, instead of the (arguably more logical) `null` value to indicate a lack of location fix?  (You would obviously need to change the return type to be an `Integer` object rather than an `int` primitive.)
2. Formatting.  Yuck!
3. Why not just return from the if branch; do away with the else branch, and return inline instead of having a `fromEnd` variable?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/LiveRideOverlay.java
```diff
@@ -172,15 +238,123 @@ private String speed() {
...
+      return null;
+    return new GeoPoint(location);
+  }
+
+  private void drawRemainingJourney(Canvas canvas, Rect box) {
+    String strRemDistance;
+    String strRemTime;
+    String ETA = "";
+    int ETAWidth=0;
+    int ETATextWidth=0;
```
oliverlockwood, line 266: Please take more care with your formatting generally - for example, we should always have spaces around `=` signs.

libraries/cyclestreets-view/src/main/res/values/strings.xml
```diff
@@ -137,5 +137,9 @@
   <string name="perm_justification_read_contacts" translatable="false"><![CDATA[
     <li>&nbsp; allow you to find the addresses of your contacts on the map, and create routes to these locations</li>
   ]]></string>
+  <string name="show_remaining_time">Show remaining time</string>
+  <string name="show_ETA">Show estimated time of arrival</string>
+  <string name="remaining">Remaining</string>
+  <string name="ETA">/ ETA</string>
```
oliverlockwood, line 143: Perhaps a better string identifier would be "or_ETA" given that you include a slash in the text value.

mvl22:
Thanks, Hilary!

I agree with the suggestions made by Oliver. I'm a totally stickler for consistency in formatting - it's something that really pays back when you develop a codebase over years! So even though maybe those might seem fussy now, I'm sure you'll be glad in the future if you ever need to come back to the code. Thanks again - great to see this new feature going in! :)

HilaryN:
Thanks - I'll take a look through and make the changes.  The single-figure hour was because I was being miserly with space - I'll take a deep breath and be more generous ;-)

oliverlockwood:
Cool, thanks.

If single-figure hour was to save space, then it would in any case have been a bit hard on anyone arriving between 11:00 and 12:59!

libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
HilaryN: I don't know if you saw my email about this - what would be the best way to do this?  Should I create a separate class or is there a better place to put it which is accessible to ControllerOverlay?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Segment.java
HilaryN: I think I was using the existing name, but I agree, cumulativeTime would be clearer, so I'll look to change the name in the existing code as well.

HilaryN:
> 
> Hi @HilaryN, I've made a number of inline comments that need addressing. I haven't gone into the depths of the logic in LiveRideOverlay, considering that I'll wait for you to address the first round of feedback first.
> 
Thanks for taking the time to look through this - much appreciated.
It may be Thurs before I can get back to you with the changes but also I'm having a re-think about the LiveRide settings (I'll probably put my re-thoughts together on Thurs as well) so, yes, do hold off looking at the logic in LiveRideOverlay.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/tiles/TileSource.java
HilaryN: I've renamed it attributionUpShift :-)

libraries/cyclestreets-view/src/main/res/values/strings.xml
HilaryN: It's not really an "or", it's a separator from "Remaining".  I could use "&", but I think "/" is neater.  So I've changed the name to "slash_ETA".

HilaryN:
LiveRideOverlay is still WIP - just your suggested changes in this commit.

HilaryN:
[Test remaining journey with adjusted text size.pdf](https://github.com/cyclestreets/android/files/4771413/Test.remaining.journey.with.adjusted.text.size.pdf)

jezhiggins:
Looking good. Going to try and get out and about with it this weekend.

HilaryN:
That would be great if you can and give some feedback on text size and anything else.

I was using it last week and sending a friend screen shots with the ETA when riding to meet up. 

Would be handy to have a "Share" button, but that can be for the future...

HilaryN:
There's a slight issue where the Active Segment doesn't always get set correctly.  In the emulator it happens when you skip a junction, i.e. go from a long way before a junction, skipping the "Get ready to turn..." and "Turn..." and going straight to a place beyond the junction.
This can cause the remaining journey to be wrong as it thinks you are a long way from the end of what it thinks should be the active segment.  But it generally goes into a Replan and corrects itself.
Hopefully it won't happen so much on a real bike unless you are in a place without signal.

HilaryN:
I got native-speaking friends to quickly check all the translations except Russian as my Russian friend was too busy.  Would benefit from more native-speaking reviewers though.  I don't speak any Turkish and wouldn't be able to sense-check a Turkish translation so I've left that in English.  (I've hardly any Russian either so that translation may be bad.)

HilaryN:
I've repeated the tests in the pdf above after splitting out the drawing of the speed and remaining journey into a separate class.
I've also checked the DistanceUntilTurn still works (pdf attached) and checked on my phone that the speed still works as I've removed a couple of calls to lastLocation().

[Test DistanceUntilTurn after calls to lastLocation() removed.pdf](https://github.com/cyclestreets/android/files/4988353/Test.DistanceUntilTurn.after.calls.to.lastLocation.removed.pdf)

oliverlockwood:
FYI @HilaryN I went and made some changes myself particularly to the new `LiveRideJourney` class - mostly in terms of naming, capitalisation, formatting etc, but there are some simplifications e.g. chaining of multiple `String.replace()` operations, and one optimization (don't calculate `remTime` if it's not necessary).

None of these changes should be functional so I haven't gone and done any testing.

With these changes in place I'm happy to take the PR in and merge it; then we can see how it fairs in beta-land!

Thank you very much for making this significant contribution - apologies it's taken me a bit of time to get round to processing it.

oliverlockwood (approved):

HilaryN:
Brilliant!
Thanks very much for cleaning that up.  Noted about the .replace and optimization.

PR #412, opened 2020-05-30
402: use vector graphics (where possible) for Photomap icons

oliverlockwood:
Related to #402.

Using vector graphics allows things to be much crisper.

As before, I've kept the original SVGs for reference, but outside what gets built into the APK.
I've imported the SVGs into Android vector XML where possible.  Unfortunately the "closure" icons make use of some SVG capability that isn't supported in Android vector XML (the latter supports only a subset of the former) - it appears as solid black, just like @si-the-pie's compass did when I tried using the vector version of that without turning it into a PNG.

I haven't dug into which particular capabilities aren't supported, but it's something that @mvl22 and/or @si-the-pie may wish to give thought to (if you want to reissue a new android-friendly SVG of the three "closure" icon that I can import, please let me know).  For now I've just turned that into a selection of PNG files giving equivalent DPI.

Lots of examples follow.
With the evidence from these, I concluded the 17dp height vectors (i.e. as close to previous size as possible) was the right way to go, and that's how this PR stands.  Feel free to weigh in.
@mvl22 @jezhiggins @si-the-pie @HilaryN your review and thoughts, please?

### Before
Screen Shot 2020-05-30 at 20 01 35

### After (21dp height vectors) 
In this, everything looks pretty good to my eye; I like the deliberately slightly-larger icons, the PNG "closure" comes out slightly more blurry but it's fine.
Screen Shot 2020-05-30 at 19 47 01

### After (17dp height vectors)
Screen Shot 2020-05-30 at 20 26 07

### Before
Screen Shot 2020-05-30 at 20 06 56

### After (21dp height vectors)
In this, the icons all look nice and crisp (I haven't shown an area with a "closure", albeit a bit large.
Screen Shot 2020-05-30 at 19 43 28

### After (17dp height vectors)
Screen Shot 2020-05-30 at 20 23 27

### Before
Screen Shot 2020-05-30 at 20 08 45

### After (21dp height vectors)
Improved size, and most are crisper.  Some (e.g. "signs", "routesigns") weirdly, are not.  I can't explain that (maybe it's the phone? maybe the API version?) but since the blurry ones are no worse than they are before the change, I don't really care.
Screen Shot 2020-05-30 at 20 10 20

### After (17dp height vectors)
Screen Shot 2020-05-30 at 20 16 15


HilaryN:
Definitely much clearer.  I don't have a strong view either way on 17 or 21dp.

mvl22:
I would favour the slightly smaller size - in some areas, there can be very high density of icons, which crowds out the map background.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/PhotoMarkers.kt
```diff
@@ -18,36 +19,29 @@ private const val NEUTRAL = "neutral"
...
+
+        val key = "pm_${category}_${metaCategory}"
 
         if (!markers.containsKey(key)) {
             try {
-                val asset = res.assets.open(key)
-                val bmp = BitmapFactory.decodeStream(asset)
-                val marker = BitmapDrawable(res, bmp)
-                asset.close()
+                val resId = res.getIdentifier(key, "drawable", res.getResourcePackageName(R.drawable.pm_bicycles_bad))
```
jezhiggins, line 41: Can we hoist the res.getResourcePackageName call out of here? Make it a val member perhaps?

jezhiggins: Otherwise, this is all lovely.

oliverlockwood: Done & done.  Good call.

PR #414, opened 2020-06-01
402: Sort out the POI Icons

### LINKED ISSUE #402: Update icon set for POIs and Photomap

oliverlockwood:
Closes #402.

In this PR, we move away from dynamically retrieving POI category icons, to embedding them with the app.  This allows for:
* better scaling of icons for different densities (including one case - "Beds for Cyclists" where the icon is completely different for <=32px and >32px, so dynamic scaling just wouldn't cut it)
* us to easily achieve a consistent dp sizing, without having to hack around with bitmaps.

In order that we don't make it hard for ourselves in future, as and when the intended POI icon set changes, I have added a Python script which:
* can be manually run to update the embedded icon set
* runs during CI, as a check (and fails the build early if the icon set needs updating).

The following screen shots show POI category list view and POIs-on-map view, for two emulators: the Pixel 2 (mirroring my phone) and the Moto G6 Play (on which I had previously reproduced the massive-icon issue highlighted by @mvl22).

## Pixel 2
Screen Shot 2020-06-03 at 06 57 34 Screen Shot 2020-06-03 at 06 57 46

## Moto G6 Play
Screen Shot 2020-06-03 at 06 56 59 Screen Shot 2020-06-03 at 06 57 14


oliverlockwood:
> Python2? What are we? Savages?

@jezhiggins I know... it made me weep to have to back-port the python 3 code I had written! 😢 

For context:
- Android builds are only supported on "Trusty" (Ubuntu 14.04) build environment
- It's impossible to get a version of Python3 later than 3.4.x (can't remember the exact patch version) on Trusty boxes
- This version suffers some SSL weirdness when you try to use the `requests` package.

oliverlockwood:
@jezhiggins now that it's ready, would you please review this one for me when you have the chance?

mvl22:
Those screens look great now :)

PR #417, opened 2020-06-05
387 offline mappacks

jezhiggins:
Downloading and managing map packs within the main app.


libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/Map.kt
```diff
@@ -0,0 +1,8 @@
+package net.cyclestreets.api
+
+data class Map(
+        val id: String,
```
oliverlockwood, line 4: picky, but 8 characters indentation seems excessive.

Also I wonder if we would be better off naming this class differently - e.g. `MapPack` - so we don't confuse it with the native `Map` class in Kotlin and/or Java.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/geojson/MapsFactory.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.geojson.Feature
+import org.geojson.FeatureCollection
+
+class MapsFactory {
+    companion object {
+        fun toMaps(featureCollection: FeatureCollection): Maps {
+            return Maps(
+                featureCollection.features
+                        .map { f -> toMap(f) }
+                        .filter { m -> isBritainOrIreland(m) }
```
oliverlockwood, line 15: I agree with this restriction as a starting point.  We can consider adding others subsequently as part of #46.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/MapPack.kt
```diff
@@ -0,0 +1,63 @@
...
+import net.cyclestreets.api.Map
+import java.io.File
+import java.io.FileInputStream
+import java.io.FilenameFilter
+import java.io.IOException
+import java.util.*
+
+class MapPack private constructor(
+        private val map: Map
+) {
```
oliverlockwood, line 16: I think this would read easier as a one-liner rather than these 3 lines, but maybe that's just me.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/Maps.kt
```diff
@@ -0,0 +1,44 @@
+package net.cyclestreets.api
+
+import android.os.AsyncTask
+
+class Maps(
+        private val packs: Collection<Map>
+): Iterable<Map> {
```
oliverlockwood, line 7: I think this would read easier as a one-liner rather than these 3 lines, but maybe that's just me.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/Map.kt
oliverlockwood: I now see it can't be `MapPack` as that would conflict.
What about `MapReference` or something similar; as effectively it acts as a reference to a downloadable map pack?

oliverlockwood:
I see this is still in partial state at present (clearly some more work to be done in terms of actually getting hold of the map pack, etc!) but hopefully my comments may still have some value.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/Map.kt
jezhiggins: I started with MapPack, realised, changed to Map as a holding name, now considering DownloadableMap

jezhiggins:
It doesn't look like much, but this a rendering a vector map that the app has downloaded from the CycleStreets server, using the Android DownloadManager.

There's still some more work to do on the UI elements - should probably ask if it's ok to download when not on Wifi, a way to uninstall them, etc, and on background updating - but moving in the right direction.

oliverlockwood:
Good work @jezhiggins! :raised_hands:
Let me know when you'd like me to do some code reviewing and I'll jump on it.

PR #419, opened 2020-06-06
95 show progress and remaining ride length during live ride 200606

HilaryN:
Ready for review of LiveRideOverlay.  This now checks for overlap between speed and remaining journey and reduces text size if there is an overlap.


oliverlockwood:
@HilaryN why have you raised a new PR instead of updating the existing PR #410?

This makes absolutely no sense to me.

Normal development practice is to update open PRs when feedback has been addressed or additional coding has been done.  Creating a separate PR means all context history of reviewing is lost.

Ideally please close this PR and update the other one; or at worst, close the old one with a comment linking to this one so that it's clear it's been superceded.

I don't think it makes sense for me to spend any time reviewing your code until this is in order - I don't currently know where to start.

HilaryN:
Sorry, I should have explained.  

I thought it would be easier for you to first check the changes I made in response to your comments in the other PR.

The changes to LiveRideOverlay are spread over several commits, before and after the changes to the other files which you have reviewed.

If I pushed the subsequent commits to the other PR it would be a nightmare to review LiveRideOverlay because of all the commits and if I squashed them into one then you would lose the history and might not be able to easily see the changes I made in response to your comments in the other PR.

So I was thinking you could 
1) check the changes in the other PR to check your comments have been addressed
2) check the changes to LiveRideOverlay in this PR, which has all the commits squashed into one.

I hope that makes better sense...

oliverlockwood:
@HilaryN At least now I vaguely understand what you were trying to achieve.

When looking at a PR, reviewers have the option to look at things commit-by-commit or whole-diff.  Having 2 separate PRs to review is, I'm afraid, a hindrance and not a help.
Please decide which one you want to work with, and close the other.  If your remaining PR is ready to review then mark it so rather than draft status.  I haven't got time at the moment to do any more interim reviewing guidance.

PR #427, opened 2020-06-16
Check route number in list before adding to history for autocomplete.

### LINKED ISSUE #93: Find route autocomplete repeats earlier items
Type a route number into the 'Open by route number' dialog and tap 'Find route'.

Then do it again for the same number and when part of the number is entered the suggestion appears. Accept the suggestion.

Repeat - but this time note that two copies of the same number appear in the suggestions.

HilaryN:
Change textAppearance so items in list are now visible. 
Fixes #93

### DIFF
--- a/libraries/cyclestreets-view/src/main/res/layout/texthistory.xml
+++ b/libraries/cyclestreets-view/src/main/res/layout/texthistory.xml
@@ -3,6 +3,6 @@
   android:id="@android:id/text1"
   android:layout_width="fill_parent"
   android:layout_height="?android:listPreferredItemHeight"
-  android:textAppearance="?android:textAppearanceMediumInverse"
+  android:textAppearance="?android:textAppearanceMedium"
   android:singleLine="true"
   style="?android:dropDownItemStyle" />
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/EditTextHistory.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/EditTextHistory.java
@@ -45,9 +45,12 @@ public class EditTextHistory extends ArrayAdapter<String> {
       lastWritten = 0;
 
     final SharedPreferences.Editor edit = prefs_.edit();
-    edit.putString(Integer.toString(lastWritten), n);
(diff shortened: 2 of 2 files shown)


HilaryN:
Autocomplete items now visible and only appear once in list, even if you go back and select item several times.

Screenshot_1592339344

libraries/cyclestreets-view/src/main/res/layout/texthistory.xml
```diff
@@ -3,6 +3,6 @@
   android:id="@android:id/text1"
   android:layout_width="fill_parent"
   android:layout_height="?android:listPreferredItemHeight"
-  android:textAppearance="?android:textAppearanceMediumInverse"
+  android:textAppearance="?android:textAppearanceMedium"
```
oliverlockwood, line 6: @HilaryN  how is this change related?  I assume this changes the colour in some way but I'm curious because I don't see any mention of that in the issue description or the PR header comment.

oliverlockwood (approved):
Approved, but one question about the change in text colour

libraries/cyclestreets-view/src/main/res/layout/texthistory.xml
HilaryN: Apologies for not making that clear.
It's a separate bug fix which I tagged on here as it's tiny.
It shows up when you go through the steps Si mentions in #93. In the suggestions dropdown, the items in the list are invisible as they are white text on a white background.  They only appear when you tap on the list and the text/background change colour.
In this set of screenshots the first one is typing the route number in first time. I then accept it.
In the second, you can see (with difficulty) that as I type, a white dropdown has appeared (with one item).  But it's white on white so you can't see the actual item.  The last screenshot is when I touch the item and the background colour changes to green:

Open route dropdown 3

This set of screenshots shows the dropdown with 2 items in the list- the second and third show the list with the first and second items selected:

Open route dropdown 4

Maybe it's an issue which appeared subsequent to the other problem being logged but as it was impeding me doing testing I thought I'd fix it at the same time.

oliverlockwood:
All is clear.  Let's get it merged!

PR #431, opened 2020-07-21
Add overlay to mark Find Place on map

### LINKED ISSUE #428: Arrow on "Find a Place"
It's great having the Find a Place, but I often find myself staring at the screen for ages trying to find the road, and usually have to go to Google Maps to find it.
It would be good to have an arrow or some indicator showing where it is.

HilaryN:
Fixes #428 

I used the "There" marker from the location editor (the red cross) for the Find Place marker.

Tested the following (see attached pdf for more detail):
Journey Planner
Photo map
Move screen
Zoom in and out
Combine with POI's, Waymarks, route highlight, location, various combinations
Back button (for the various combinations above)
Follow location
[Test FInd Place marker 200721.pdf](https://github.com/cyclestreets/android/files/4956019/Test.FInd.Place.marker.200721.pdf)

Screenshot with the marker:

Screenshot_1594731460

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/FindPlaceOverlay.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/FindPlaceOverlay.kt
@@ -17,26 +17,26 @@ import org.osmdroid.views.overlay.Overlay
 
 class FindPlaceOverlay(context: Context, cycleMapView: CycleMapView) : Overlay(), Undoable {
     // (Passing CMV as parm so centreOn can be accessed, which we need to calc screen coords)
-    private val placeMarker_: Drawable?
-    private var cMapView_: CycleMapView
+    private val placeMarker: Drawable
+    private var cMapView: CycleMapView
     private var halfWidth: Int = 0
     private var halfHeight: Int = 0
 
     private val controller = OverlayHelper(cycleMapView).controller()
 
     init {
-        val res: Resources = context.getResources()
-        placeMarker_ = ResourcesCompat.getDrawable(res, R.drawable.x_marks_spot, null)
-        if (placeMarker_ != null) {
-            halfWidth = placeMarker_.intrinsicWidth / 2
-            halfHeight = placeMarker_.intrinsicHeight / 2
-        }
-        cMapView_ = cycleMapView
+        val res: Resources = context.resources
+        placeMarker = ResourcesCompat.getDrawable(res, R.drawable.x_marks_spot, null)!!
+
+        halfWidth = placeMarker.intrinsicWidth / 2
+        halfHeight = placeMarker.intrinsicHeight / 2
+
+        cMapView = cycleMapView
     }
 
     override fun draw(canvas: Canvas, mapView: MapView, shadow: Boolean) {
-        val foundPlace: IGeoPoint? = cMapView_.getFoundPlace()
-        if ((foundPlace == null) || (placeMarker_ == null)) return
+        val foundPlace: IGeoPoint? = cMapView.getFoundPlace()
+        if (foundPlace == null) return
 
         if (!controller.checkUndo(this))
             controller.pushUndo(this)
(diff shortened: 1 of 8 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/FindPlaceOverlay.kt
```diff
@@ -0,0 +1,66 @@
...
+    private var cMapView_: CycleMapView
+    private var halfWidth: Int = 0
+    private var halfHeight: Int = 0
+
+    private val controller = OverlayHelper(cycleMapView).controller()
+
+    init {
+        val res: Resources = context.getResources()
+        placeMarker_ = ResourcesCompat.getDrawable(res, R.drawable.x_marks_spot, null)
+        if (placeMarker_ != null) {
```
oliverlockwood, line 30: Do we really need this `if` protection?

Elsewhere we make Drawables from static resources be non-nullable (i.e. type of `Drawable` instead of `Drawable?`) and force non-nullability at the time of instantiation.  e.g. in `RotateMapOverlay.kt`, we do like:
```kt
private val onIcon: Drawable

init {
    onIcon = ResourcesCompat.getDrawable(context.resources, R.drawable.compass, null)!!
}
```

I think it would make the code simpler to take this approach here as well.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/FindPlaceOverlay.kt
```diff
@@ -0,0 +1,66 @@
...
+            halfWidth = placeMarker_.intrinsicWidth / 2
+            halfHeight = placeMarker_.intrinsicHeight / 2
+        }
+        cMapView_ = cycleMapView
+    }
+
+    override fun draw(canvas: Canvas, mapView: MapView, shadow: Boolean) {
+        val foundPlace: IGeoPoint? = cMapView_.getFoundPlace()
+        if ((foundPlace == null) || (placeMarker_ == null)) return
```
oliverlockwood, line 39: With the above, we'll be able to remove the null check on `placeMarker_` here as well.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/FindPlaceOverlay.kt
```diff
@@ -0,0 +1,66 @@
...
+import net.cyclestreets.views.CycleMapView
+import org.osmdroid.api.IGeoPoint
+import org.osmdroid.api.IProjection
+import org.osmdroid.views.MapView
+import org.osmdroid.views.overlay.Overlay
+
+class FindPlaceOverlay(context: Context, cycleMapView: CycleMapView) : Overlay(), Undoable {
+    // (Passing CMV as parm so centreOn can be accessed, which we need to calc screen coords)
+    private val placeMarker_: Drawable?
+    private var cMapView_: CycleMapView
```
oliverlockwood, line 21: My personal view (and one which I have gradually been applying across the codebase) - there's really no need for the trailing underscores in the names of  private vars / vals - I think this is a legacy from when Jez originally wrote the app and Java was younger 😄 

So I would be inclined to remove these from the names.

oliverlockwood (approved):
I'm happy with the changes; but I have added a few comments, which when addressed will make the code a little more maintainable in the long-term.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/FindPlaceOverlay.kt
HilaryN: OK, I did wonder about that convention and guessed it was from earlier days, as I see Android Studio now colours them according to scope.  Thanks for confirming that.

HilaryN:
Made changes and did a quick re-test to check Find Place marker displays in centre of screen and goes away with back button in appropriate sequence with waymarkers.

oliverlockwood:
Thanks for the efforts @HilaryN, great stuff!

PR #432, opened 2020-07-22
227 make url in poi bubbles clickable short text

### LINKED ISSUE #227: Change website link displayed in POI bubbles to be clickable short text
... instead of a non-clickable, potentially verbose, full URL.

As suggested by @mvl22 when reviewing #224.

HilaryN:
Fixes #227 

See issue 227 for screenshots


oliverlockwood:
I'm not a UI design hero, but it strikes me as potentially counterintiuitive that:
- tapping some bubbles closes them
- tapping other bubbles opens a website, and then on returning to the app the bubble is disappeared.

I would be more in favour of a clickable link _within_ the bubble then leaving the rest of the bubble behaviour consistent regardless of the presence of a website.

Any other opinions @mvl22, @jezhiggins ?

mvl22:
> I would be more in favour of a clickable link within the bubble then leaving the rest of the bubble behaviour consistent regardless of the presence of a website.

Yes, UI consistency is a very strong principle that shouldn't be compromised.

I think it needs an icon (maybe a world icon) that represents the link. Or just make the text of the establishment linked.

HilaryN:
Thanks for the feedback.

I'm a bit confused as the behaviour is the same with or without a link - tapping on the bubble closes the bubble.  If there is a website it additionally goes to the website.

Do you mean, only go to the website if they click on the clickable link (and make that look like a hyperlink or an icon) within the bubble, and leave the bubble open?

I could (try and) make the text appear like a hyperlink, but with it being so small my worry is that it would quite difficult to tap in the exact spot.

HilaryN:
I'm mulling over some possibilities:

- change the behaviour so that clicking _outside_ the bubble closes it, and clicking inside the bubble takes you to the website if there is one. (A friend of mine claims to have "fat fingers", not that you would know to look at them, and has problems even typing messages.)

- change wording to "Tap here for more details"

-  make "Tap here for more details" look like a link.

HilaryN:
Feedback pleeease :-) 

- What's your preference - A, B, C or D?  (Feel free to suggest something different.)

- Is your preference definitely to tap on the link to show the website rather than clicking anywhere within the bubble?

I think my preference is for C.  My worry about D is that it could just look like a heading rather than a link.
I'll have a go at making it so you have to tap on the link and ask friends to try it out.

First 2 are Pixel 3 emulator, second 2 are Nexus One emulator, hence the different-sized font.

B IS underlined, though it doesn't look much like it in the image.

I've done the link text specifically as bold / underline as I couldn't find a specific "link" style for Paint.  I could do it as a clickable TextView, though I like it this way.  Let me know if you have any thoughts.

POI link screenshots

mvl22:
D is clearest and most intuitive to me.

I'm never in favour of UIs that add "click here" type of text when you can simply underline or use an icon.

oliverlockwood:
I'm also inclined to prefer D.

Ping me when you've updated the PR to this effect, @HilaryN, and I'll do my best to review it promptly.

HilaryN:
Currently the bubble disappears when you pause the app (go to the Home screen or another app) or if you rotate the screen.  When doing the Find Place marker I noticed an issue where RouteMapFragment is destroyed in the MainNavDrawerActivity.**onResume**() - in other words the fragment is being destroyed as the app is resuming, and I don't understand why.

I could do a workaround, as I have done with the Find Place marker, but I don't really like to do this as it goes against the Android guidelines and it makes the marker (or in this case would make the bubble) more permanent than I would wish.  I'd like the Find Place marker / POI bubble to stay on the screen if the app is paused (i.e. the user is away for just a short time), but go away if the app is destroyed.  Using the ViewModel seems to be the correct way to do this – but again this won't work because of the fragment being destroyed.

 I'd like to find out why the fragment is being destroyed and will carry on looking into that.  I may report it as a bug, as I can't find anything in the Android documentation about it.

HilaryN:
[Test POI link.pdf](https://github.com/cyclestreets/android/files/5069430/Test.POI.link.pdf)

oliverlockwood:
@HilaryN the PDF demonstration of the behaviour looks good.  I'll review the code now.

libraries/cyclestreets-view/src/main/res/values-tr/strings.xml
```diff
@@ -63,6 +63,8 @@
   <string name="report_title">Raporunuz</string>
   <string name="report_location_hint">Neredeki konumda raporu taımladın?</string>
 
+  <string name="tap_here">Tap here to show website</string>
```
oliverlockwood, line 66: Given that the English text is now `More details...`, I suspect that the other language files should show the same if we haven't translated this string yet.

libraries/cyclestreets-view/src/main/res/values-fr/strings.xml
```diff
@@ -63,6 +63,8 @@
   <string name="report_title">Votre rapport</string>
   <string name="report_location_hint">Où est l\'emplacement de votre rapport décrit?</string>
 
+  <string name="tap_here">Appuyez ici pour voir le site Web</string>
```
oliverlockwood, line 66: All these translations need updating please

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Brush.java
```diff
@@ -52,5 +52,18 @@ public static Paint createTextBrush(final int size, final int r, final int g, fi
     return paint;
   }
 
+  public static Paint createUrlBrush(final int size) {
+    final Paint paint = White;
+
+    paint.setTextAlign(Paint.Align.CENTER);
+    paint.setTypeface(Typeface.DEFAULT);
+    paint.setTextSize(size * 2);
```
oliverlockwood, line 60: These 5 lines could be replaced with `createTextBrush(size)`

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Draw.java
```diff
@@ -10,43 +10,76 @@
...
+      ascent = fm.ascent;
+      boxHeight = (lineHeight * lines.length) + doubleOffset;
+    }
 
     bounds.left = pos.x - (width/2);
     bounds.right = bounds.left + width;
+    // Subtract to go UP
     bounds.top = pos.y - (boxHeight + (doubleOffset * 2));
+    // Add to go DOWN
     bounds.bottom = bounds.top + boxHeight;
```
oliverlockwood, line 60: useful comments 👍

oliverlockwood (approved):
Some comments which it would be good to address, but generally I'm happy with this.  Nice work, apologies it took a while for me to get round to reviewing it.

libraries/cyclestreets-view/src/main/res/values-fr/strings.xml
HilaryN: Done - I was holding off in case of any change to the text :-)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Brush.java
HilaryN: Done, and I did a quick re-test to make sure the link appears OK in the bubble.  
I accidentally missed the file off the commit, hence the extra commit in addition to the translations - sorry!  (Let me know if you want them squashing.)

HilaryN:
There is still the outstanding issue that the bubble disappears when you pause / resume, but I'll raise that as a separate issue for discussion as it affects other areas as well.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Brush.java
oliverlockwood: No worries.

No need to worry about squashing your commits to the PR; we squash the whole PR into one commit as part of merging.

cyclestreets.app/src/main/assets/whatsnew.html
```diff
@@ -8,13 +8,17 @@
 <h3>Release 3.9 (BETA)</h3>
 <strong>Features:</strong>
 <ul>
-  <li>Coming soon...</li>
+  <li>POI URLs (e.g. bike shop websites) now appear as clickable links.</li>
```
oliverlockwood, line 11: FYI @HilaryN I pushed an extra commit here to update the What's New (both the in-app page, this file, and the Beta download store listing, in `beta.txt`, to describe the new functionality.

PR #437, opened 2020-09-10
Rip out vector mapfile setting

### LINKED ISSUE #436: App crash accessing settings since update to Android 10
Yesterday my Fairphone 3 was updated to Android 10. Since then opening the left sidebar and tapping settings causes the app to crash.

Prior to the OS update this wasn't an issue.

Version 3.9.0 on the Google play app store. I have beta updates enabled.

I've tried clearing the cache and app data, both of which have the same crash.

jezhiggins:
Closes #436.

It's causing crashes on some phones, and will be replaced with a new
mechanism soon, see PR #417.

### DIFF
--- a/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/SettingsFragment.kt
+++ b/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/SettingsFragment.kt
@@ -22,7 +22,6 @@ private const val PREFERENCE_SCREEN_ARG: String = "preferenceScreenArg"
 private val SETTINGS_ICONS = mapOf(
     "screen-maps-display" to GoogleMaterial.Icon.gmd_map,
     "mapstyle" to null,
-    "mapfile" to null,
     "confirm-new-route" to null,
     "screen-routing-preferences" to GoogleMaterial.Icon.gmd_directions,
     "routetype" to null,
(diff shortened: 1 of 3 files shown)


libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/SettingsFragment.kt
```diff
@@ -165,37 +143,9 @@ class SettingsFragment : PreferenceFragmentCompat(), SharedPreferences.OnSharedP
...
-        var index = mapfilePref.findIndexOfValue(mapfile)
-        if (index == -1)
-            index = 0 // default to something
-
-        mapfilePref.setValueIndex(index)
-        mapfilePref.summary = mapfilePref.entries[index]
-    }
-
-    private fun setAccountSummary() {
+   private fun setAccountSummary() {
```
oliverlockwood, line 148: I think you've accidentally de-indented this by 1 character 😂 

I clearly don't care enough to request a change, but you may feel like fixing it as part of your broader PR.

oliverlockwood:
@jezhiggins the only possible issue I can see is: what if someone already has vector maps selected as their preference, and then they upgrade to this version?  What will happen then?

jezhiggins:
It should fallback to the default renderer. In fact it will have already done that as it wouldn't be able to open the map file.

oliverlockwood (approved):

PR #438, opened 2020-09-17
Allow background location access

### LINKED ISSUE #429: Enable voice navigation even when screen timed out
User request:

"recently discovered your awesome android app! A small q - do you know how to make the voice navigation keep working even if the screen timesout? If I force the screen on it works - but then drains battery quite quickly."

https://twitter.com/sim0nw/status/1273872031864500229

HilaryN:
Closes #429.

### DIFF
--- a/libraries/cyclestreets-view/src/main/AndroidManifest.xml
+++ b/libraries/cyclestreets-view/src/main/AndroidManifest.xml
@@ -23,6 +23,7 @@
               android:launchMode="singleTop"
               android:screenOrientation="portrait" />
 
-    <service android:name="net.cyclestreets.liveride.LiveRideService" />
+    <service  android:name="net.cyclestreets.liveride.LiveRideService"
+              android:foregroundServiceType="location" />
   </application>
 </manifest>
--- a/libraries/cyclestreets-view/src/main/AndroidManifest.xml
+++ b/libraries/cyclestreets-view/src/main/AndroidManifest.xml
@@ -6,6 +6,7 @@
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
   <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
   <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+  <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
   <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
   <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
   <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/LiveRideActivity.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/LiveRideActivity.kt
@@ -1,11 +1,6 @@
 package net.cyclestreets
 
 import android.Manifest
-import net.cyclestreets.liveride.LiveRideService
-import net.cyclestreets.util.GPS
-import net.cyclestreets.util.MessageBox
-import net.cyclestreets.views.CycleMapView
-
 import android.app.Activity
 import android.content.ComponentName
 import android.content.Context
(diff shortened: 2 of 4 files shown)


HilaryN:
This is just a quick temporary fix which will allow an Android 11 user to change location permissions in the device settings to "Allow all the time" so that voice navigation can be used when the screen is locked.
But you might want to wait until we have the proper request for permission in place rather than approving this temporary change.

oliverlockwood:
Why don't you go ahead and do this one completely, if you're in the area? Look for usages of PermissionsKt, or something like that (from memory), or for usages of ACCESS_FINE_LOCATION, and specifically the one which fires where we launch LiveRide.

HilaryN:
Yes, I'm taking a look.  But don't let that stop you either as it may take me a while :-)
We're not allowed to ask for both foreground and background location at the same time, so it needs to be a separate request.

HilaryN:
Actually, I think I've done it now.  Just need to test and put some screenshots together.

oliverlockwood:
Good work.

FYI Incidentally I have in the back of my mind to give people a descriptive "this is why we're asking for permission X" ahead of showing the system dialog, but that's a separate topic 🙂

HilaryN:
This is still WIP and I haven't added the translations yet (or modified the text of the request as yet).

All seemed to be working as expected until I added the 
android:foregroundServiceType="location"
to the manifest.

After doing that, when I tested on Android 11, the voice navigation still worked with the screen locked, even though I selected the option to ask every time (i.e. NOT "Allow all the time").

I don't really understand the android:foregroundServiceType="location" so I'll try and find out more.

oliverlockwood:
@HilaryN that's interesting.

In April I added into `LiveRideService` the line:
```stage = LiveRideStart(this, tts).setServiceForeground(this)```
Under the covers this ends up calling
```service.startForeground(NOTIFICATION_ID, notification)```

The Javadoc for this method says, among other things:
```
     * If your service is started (running through {@link Context#startService(Intent)}), then
     * also make this service run in the foreground, supplying the ongoing
     * notification to be shown to the user while in this state.
     * By default started services are background, meaning that their process won't be given
     * foreground CPU scheduling (unless something else in that process is foreground) and,
     * if the system needs to kill them to reclaim more memory (such as to display a large page in a
     * web browser), they can be killed without too much harm.  You use
     * {@link #startForeground} if killing your service would be disruptive to the user, such as
     * if your service is performing background music playback, so the user
     * would notice if their music stopped playing.
...
```

My intention when I made this change was to give LiveRide more "priority" as I'd personally had some experience (admittedly on my old cheapo Moto G, not my new shiny Pixel 2) that LiveRide wouldn't really "get going" properly if too many other apps were open.

My recent suggestion to set `android:foregroundServiceType` came from reading https://developer.android.com/training/location/permissions#foreground.

Thinking about this further... Maybe if you set your service as foreground (as we already do) and have set the foreground service type to be location related, since that means you're _always_ displaying a notification of some sort, then we don't have to jump through the background location hoops after all?  That would be a result if so!

oliverlockwood:
Yep, reading in more detail:
>The system considers your app to be using foreground location if a feature of your app accesses the device's current location in one of the following situations:
>-  An activity that belongs to your app is visible.
>-  Your app is running a foreground service. When a foreground service is running, the system raises user awareness by showing a persistent notification. Your app retains access when it's placed in the background, such as when the user presses the Home button on their device or turns their device's display off.
>   Additionally, it's recommended that you declare a foreground service type of location, as shown in the following code snippet. On Android 10 (API level 29) and higher, you must declare this foreground service type.

This is definitely good news for us.

HilaryN:
I see (I think!)  The fact that a notification is shown means that there is no need to request background location access because the app isn't doing anything sneakily in the background even though it is running in the background.

So probably the only change needed is that foreground service type line.

jezhiggins:
Really good work. Thank you both.

mvl22:
> The fact that a notification is shown means that there is no need to request background location access because the app isn't doing anything sneakily in the background even though it is running in the background.

Yes, I guess there is a difference between an app not actually being used, i.e. genuinely in the background, and a foregrounded app that is still running while the screen has got locked.

oliverlockwood:
Closing in favour of #440.

Issue #411, opened 2020-05-25
Up-front request for photo permissions 


mvl22:
This dialog appears on first-run. It certainly appears rather intrusive.

Can we ensure that the photo permission is only requested at the point of wanting to add a photo? That is the only circumstance when that is required, and many users won't want to use the Photomap feature anyway.

![Screenshot_20200518-142929](https://user-images.githubusercontent.com/361423/82843732-fe4e7a00-9ed5-11ea-9e7a-67a3de01c84d.png)

jezhiggins:
This isn't asking for permission to take a photo, it's asking for permission to use the filesystem. Filesystem access is all or nothing, if you have the filesystem permission you can read almost everything, hence "access photos, media and files". The reason we need this permission is to cache map tiles. If that permission was denied, we probably could operate (although I've never tried) but we'd constantly be hitting the network. 

We probably can delay asking so it doesn't overlay the splash screen, but we do need this permission and we do need it pretty promptly.

oliverlockwood:
I consider this closed by #440.  We're still going to ask up-front for permissions, but it should now be a lot clearer _why_ we're asking.

mvl22:
Can I just check why caching map tiles involves camera roll access?

We never want the map tile bitmaps to appear as photos - they’re just temporary memory.

oliverlockwood:
@mvl22 please reread Jez's answer to this above. It's file system access, not specifically camera roll access.

jezhiggins:
Yes, that's right - it's file system access - but the language of the permissions pop-up phrases it in terms of photos and files. The specific wording varies by some combination of device/Android version/vendor, but will always include a reference to photos.

PR #440, opened 2020-09-21
Fix permissions for LiveRide, and generally improve how we request permissions.

### LINKED ISSUE #429: Enable voice navigation even when screen timed out

oliverlockwood:
Closes #429, #411.

Supercedes #438.

The idea of this is to:
- take what @HilaryN figured out in #438 and apply that (the first commit)
- improve the "justification" screens that give context for why we're asking for permissions, by:
   - making the text clearer and more thorough
   - always including them (including before first attempt to request, unlike previously)
   - if the user has clicked "Deny" and "Never ask again" then actions which require a given permission should now give an updated justification message, and then take the user to the app settings screen directly, from where they have an opportunity to grant the permission (previously, in this scenario no justification would be seen and no opportunity to change permissions would be given, meaning the functionality - e.g. clicking LiveRide - would just do nothing).

@jezhiggins and @HilaryN, please review.

@HilaryN, is there any chance you could please do some of your thorough testing on a variety of versions, to check the behaviour is as I intended in the description above?  (I haven't got the time to test just now, but managed to squeeze in the coding herein).

Cheers.

### DIFF
--- a/libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
+++ b/libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
@@ -38,6 +38,8 @@ public class CycleStreetsPreferences
   public final static String PREF_SHOW_REMAINING_TIME = "show-remaining-time";
   public final static String PREF_SHOW_ETA = "show-ETA";
 
+  public static final String PREF_PERMISSION_REQUESTED_PREFIX = "permission-requested-";
+
   public final static String MAPSTYLE_OCM = "CycleStreets";
   public final static String MAPSTYLE_OSM = "CycleStreets-OSM";
   public final static String MAPSTYLE_OS = "CycleStreets-OS";
@@ -144,6 +146,14 @@ public class CycleStreetsPreferences
 
   public static boolean showEta() {
     return getBoolean(PREF_SHOW_ETA, true);
+  }
+
+  public static boolean permissionPreviouslyRequested(String permission) {
+    return getBoolean(PREF_PERMISSION_REQUESTED_PREFIX + permission, false);
+  }
+
+  public static void logPermissionAsRequested(String permission) {
+    putBoolean(PREF_PERMISSION_REQUESTED_PREFIX + permission, true);
   }
 
   public static boolean uploadSmallImages() {
--- a/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleMapFragment.kt
+++ b/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleMapFragment.kt
@@ -70,7 +70,7 @@ open class CycleMapFragment : Fragment(), Undoable {
     override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<out String>, grantResults: IntArray) {
         super.onRequestPermissionsResult(requestCode, permissions, grantResults)
 
-        Log.d(TAG, "Permissions granted (with 0) or denied (with -1): ${permissions.joinToString()}, ${grantResults.joinToString()}")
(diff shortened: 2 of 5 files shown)


HilaryN:
Yes, will do.

HilaryN:
Just done the first couple of tests on Adnroid 11 so far.  The second permission justification box looks good (perhaps "change this" could be "change this location permission" to be more specific?)

Seem to be having an issue with the "Ask every time" though.  After denying a few times then changing to "Ask every time", it seems to think the permission is still set to Deny.  Changing to "Allow while using this app" works OK.

When selecting "Ask every time" first time around it continues to LiveRide but doesn't ask for permission again on subsequent times using LiveRide:

[Test background location permissions 200921 Tests 1 and 2.pdf](https://github.com/cyclestreets/android/files/5255596/Test.background.location.permissions.200921.Tests.1.and.2.pdf)

HilaryN:
Just attaching a screenshot of the second box which comes up after denying, for the others to see:

Screenshot_1600695000

HilaryN:
I occasionally get slightly weird things happen which I can't then reproduce.  One example I've had a couple of times today is that CycleStreets location permission isn't appearing in the general Settings even though it has been set.  Here I had set it to ask every time but it's not appearing in the Settings (first 2 screenshots).  You can see it in the last screenshot (I didn't change it here, this is the screen which appeared after clicking OK on the permissions justification box):

CycleStreets permission not in settings

It's happening again now.  I've set it to "Allow while using the app" so LiveRide works fine, but I have no way of going in to the settings to change it to "Deny", for example.  I guess this is an Android problem rather than a problem with the app.

HilaryN:
I slightly misunderstood the "Ask every time" - it only asks again if the app is closed and opened again.  (I do this by tapping the tiles / little square button at the bottom and going to the "Clear All" to close it, then tapping on the app icon again.)

There is still a slight issue where we can't tell if the user has changed the permission to "Ask every time" when we take them to settings.  As far as the app is concerned, we only know if permission is granted or denied.  "Ask every time" returns "Permission denied".
I guess we can assume that if they tap on the LiveRide button again after denying twice then it's OK to ask them again and we can maybe clear out the logPermissionAsRequested and start again.

oliverlockwood:
After some local emulator testing (only on API 29 - I didn't have time for more), I've made some changes, including rendering out a friendly description of the permission we're requesting, and improving layout etc.

See screenshot below for an example rendering of this, and `strings.xml` for the full details for each permission type.
@HilaryN regarding the edge case you describe, I think it's probably just fine as it is - if a user keeps on trying to use a part of the app which requires an additional permission, it's reasonable to keep on giving them the option (at that point) to upgrade the permissions they've granted to the app.

Screen Shot 2020-09-24 at 16 06 46

In my view this PR is ready to go now.  @jezhiggins can you please review when you have the chance?

oliverlockwood:
P.S. Thank you @HilaryN for your testing, it has been very helpful

mvl22:
That text looks good.

You could optionally add to the end of the last paragraph:

"This can be independently verified as our app is open source."

oliverlockwood:
@mvl22 I've incorporated your suggestion.

I'm going to go ahead and merge this now.  Next steps:
- @jezhiggins please feel free to add any retrospective feedback when you get to this.
- @swattor once a few hours have passed (to allow the new beta build to be accessible on the Play store), please download it and confirm whether or not the change has fixed your issue.
- I will do some testing with my phone as it is (running Android 10) and then allow the upgrade to Android 11 which it's been nagging me about for the past few days, and then I'll test again.

PR #443, opened 2020-10-04
442: add speech fixing for St Somebody Road

### LINKED ISSUE #442: "St" for text-to-speech
User feedback:

The Android app seems to be expanding "St" incorrectly with text to speech. I don't think it did this before the changes to speaking road numbers (it, or Android, used to have trouble with "St." with a dot). It was expanding it to "Street" when I was cycling up [St Cross Road in Winchester](https://www.openstreetmap.org/way/100098754#map=16/51.0475/-1.3181&layers=CN). And yes - the signs do have "St" rather than "Saint".

oliverlockwood:
Fixes #442.

Nice and simple.  Review please 😄


jezhiggins:
I think this probably the best we can do, given the vagaries of naming things :)

jezhiggins (approved):

PR #444, opened 2020-10-04
Update compile SDK from 29 (Android 10) to 30 (Android 11)

oliverlockwood:
Based on https://developer.android.com/about/versions/11/behavior-changes-all I think this should be fine (we've already improved our approach around permissions in a relevant way).

Based on https://developer.android.com/about/versions/11/behavior-changes-11 I am **not** yet upgrading the target SDK - since I think the statement of:
> Scoped storage enforcement: Access into external storage directories is limited to an app-specific directory and specific types of media that the app has created.

has a potential impact until Jez's work on the offline mapping download has been completed.

@jezhiggins @mvl22 @HilaryN please review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,9 @@ android:
 before_install:
   - ./ciLicense.sh
   - sudo pip install requests
+  - touch $HOME/.android/repositories.cfg
+  - yes | sdkmanager "platforms;android-30"
+  - yes | sdkmanager "build-tools;30.0.2"
 
 install:
   - ./checkPoiTypeIcons.sh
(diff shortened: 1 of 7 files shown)


jezhiggins:
The scoped file access is already in hand, so there's no need to hold back because of that.

oliverlockwood:
> The scoped file access is already in hand, so there's no need to hold back because of that.

As in, already merged in, or already in your PR which is outstanding?

jezhiggins:
In the outstanding PR.

HilaryN:
Looks good to me. Not sure if it's a relevant test but I quickly checked voice nav works when screen locked for Android 11, 10 and 9 emulators.

oliverlockwood:
On that basis @jezhiggins I'll keep that change out of this PR, so we can merge this in without waiting for yours to be ready.  I've raised #445 and #446 to track the outstanding dependency upgrade bits; if you would be so kind to review as it stands, that'd be great.

jezhiggins (approved):

Issue #448, opened 2020-10-11
Unable to upload photos due to permission denied


smsm1:
On Fairphone 3 with Android 10 when trying to upload a previously taken photo I get the following error message:

![Screenshot_20201011-172949](https://user-images.githubusercontent.com/82719/95691339-5c29f700-0c16-11eb-9c3f-f788e1dd6e5c.png)

The current permissions for the CycleStreets app are:

![Screenshot_20201011-225820](https://user-images.githubusercontent.com/82719/95691348-69df7c80-0c16-11eb-9304-b68e3c60acb8.png)

So access to storage seems to be allowed for the app.

I'm suspecting this is related to an Android 10 change.

oliverlockwood:
Thanks for raising, @smsm1. I'll investigate. Definitely an Android change, sigh.

HilaryN:
I think it's failing on the Exif access in AddPhotoFragment - if I comment out those lines it doesn't bring up that message.

https://developer.android.com/training/data-storage/shared/media#media-location-permission

oliverlockwood:
Good analysis @HilaryN. If you'd like to fix this by adding the permission to the manifest and using the PermissionsKt method to request at the appropriate button-click point in runtime, feel free 😉 or I'll get to it probably on Sunday at a guess.

HilaryN:
Unfortunately this permission isn't working in the way I'd expect.  If permission has already been granted for READ_EXTERNAL_STORAGE, then when you request permission for ACCESS_MEDIA_LOCATION no permission request box pops up... but then the permission is granted!

If you don't request permission then it fails with Permission denied, as highlighted by smsm1.

It seems silly (and confusing to the user) to pop up a justification box explaining we're about to ask for permission to get the location from their photo and then no dialogue box appears asking for permission.   We could just stick in the requestPermission without a justification message so the permission is granted in the background but that seems silly as well - and if Google change the behaviour in the future then we'd need to change this again.

I'm still trying to get my head around what would be the best way to do it but in the meantime I've gone for the simpler option of using the requestLegacyExternalStorage="true" as a temporary solution.

The app will need reviewing for scoped storage in Android 11 anyway so maybe the ACCESS_MEDIA_LOCATION permission can be done at the same time... :-)

HilaryN:
@oliverlockwood this can be closed now, can't it?

HilaryN:
On second thoughts, maybe leave open until more permanent fix.

PR #449, opened 2020-10-20
Added legacy storage flag

HilaryN:
The documentation says this is only valid for Android 10, but I've tested it on Android 11 and the exif permission still works on there (though it worked on there anyway without this change).
#448 
[Test legacy permissions flag.pdf](https://github.com/cyclestreets/android/files/5410922/Test.legacy.permissions.flag.pdf)


oliverlockwood:
This change is already made as part of #453, which I'm about to merge, so I think I'll close this PR now as redundant.

Issue #451, opened 2020-11-24
"Ask every time" permission doesn't work after denying twice


HilaryN:
On Android 11, if you deny location permission twice then go to settings and set permission to "Ask every time", the app thinks permission is still denied and LiveRide / Current location won't work.

I know it's a minor issue, but while it was fresh in my mind I thought I'd fix it.

PR coming soon. :-)

mvl22:
Sounds like this can be closed, once it has been merged in? I see Jez has reviewed.

HilaryN:
Merged.

Issue #452, opened 2020-11-24
Have to tap LiveRide button again after granting location permission


HilaryN:
If you tap the LiveRide or location button and the permission request comes up, and you grant permission, you then have to tap the LiveRide or location button again for it to work.

Probably affects me more than the average user as I am frequently reinstalling, but I thought I'd fix it for a better (new) user experience :-)

PR coming up...

mvl22:
Sounds like this can be closed, once it has been merged in? I see Jez has reviewed.

HilaryN:
Merged.

PR #453, opened 2020-11-24
Changes to location permission: allow Ask Every Time after 2 x Deny. …

HilaryN:
… Move action to callback so you don't have to press LiveRide or location button twice.

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.kt
@@ -10,11 +10,11 @@ import android.net.Uri
 import android.os.Build
 import android.provider.Settings
 import android.util.Log
+import android.widget.Toast
 import androidx.annotation.RequiresApi
 import androidx.core.app.ActivityCompat.startActivity
 import androidx.fragment.app.Fragment
-import net.cyclestreets.CycleStreetsPreferences.logPermissionAsRequested
-import net.cyclestreets.CycleStreetsPreferences.permissionPreviouslyRequested
+import net.cyclestreets.CycleStreetsPreferences.*
 import net.cyclestreets.util.Permissions.justifications
 import net.cyclestreets.view.R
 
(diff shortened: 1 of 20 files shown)


HilaryN:
Fixes #451 and #452 .

Summary of changes

Fixed issue of “Ask Every time” not working after permission denied twice.
For LiveRide and Current Location, execute these immediately after permission granted rather than user having to tap button again.  (Ideally I would have liked to have done this for all the actions triggering a permission request, but due to a pause/resume issue I decided not to do workarounds at this stage as I am hoping to look into the pause/resume issue.)
Combined doOrRequestPermissions for activity and fragment to save on duplication of code.

Notes on individual file changes

Permissions.kt

Added flag of SettingsLastTime in order to make the “Ask every time” work after user has denied twice.
Added function requestPermissionResultAction to be used in the permission request callbacks to set the flags depending on whether permission granted or not and to carry out the appropriate action triggering the permission request.
Added request code as parm to doOrRequestPermissions in order to be able to distinguish between location permission request for LiveRide or Follow Location, so that the appropriate action (start LiveRide or Enable/Follow Location) can be taken immediately from the permission callback if permission granted (rather than user having to tap button again).
Combined the doOrRequestPermissions and requestPermissions functions for fragment and activity.  (Changed the calls in the various fragments and activity to reflect the extra parms.)

Added Toast message to appear if user tries to access function for which permission is denied.

CycleMapFragment
Added fragment as parm when initialising CycleMapView (because parm added to CycleMapView / LocationOverlay)
Added parms to doOrRequestPermissions: activity (null), (as doOrReq combined for activity and fragment) and request code.
Called requestPermissionsResultAction to set flags and carry out action if permission granted.


RouteMapFragment
Added parms to doOrRequestPermissions: activity (null), (as doOrReq combined for activity and fragment) and request code so that callback can determine which action needed.
Added onRequestPermissionsResult callback to update flags and call LiveRide or enable Current Location if permission granted.  Request code determines which action required if permission granted.

AddPhotoFragment
Added fragment (“this”) as parm when initialising CycleMapView (to be passed to LocationOverlay so it will request permission on this fragment and callback in this fragment will be triggered)
Added parms to doOrRequestPermissions: activity (null), (as doOrReq combined for activity and fragment) and request code.
Added onRequestPermissionsResult callback to update flags.

LiveRideActivity
Added fragment parm (=null) when initialising CycleMapView

CycleStreetsConstants
Added constants for request codes.

strings x 8

LocationEditorActivity:
Added onRequestPermissionsResult callback to update flags and enable Current Location if permission granted.
Added parm when initialising CycleMapView (so that LocationOverlay can distinguish calling class when calling doOrRequestPermissions)

CycleMapView
Added fragment parm to constructor.
Added fragment parm when initialising LocationOverlay (so that LocationOverlay can call doOrRequestPermissions with appropriate parms)
Added method to save location preferences so that these can be reinstated in onResume.


LocationOverlay
Added fragment as parm to constructor (so that doOrRequestPermission can be called with the appropriate parms and later the appropriate callback will be triggered)
Added request code parm to doOrRequestPermission calls so that callback can determine action if permission granted.
doOrRequestPermission is called with the appropriate parms depending on whether CycleMapView / LocationOverlay has been called from a fragment or activity.
(CycleMapView / LocationOverlay is called in LocationEditorActivity, LiveRideActivity, CycleMapFragment, AddPhotoFragment, so these have been updated with the extra parm.)

PlaceViewBase
Added parms to doOrRequestPermissions: fragment (null), (as doOrReq combined for activity and fragment) and request code.

CycleStreetsPreferences
Added extra flag (SettingsLastTime) and a function to clearPermissionRequested

[Test permissions after Deny.pdf](https://github.com/cyclestreets/android/files/5592631/Test.permissions.after.Deny.pdf)

Toast message when permission denied

jezhiggins (approved):

mvl22:
Thanks Hilary!

oliverlockwood:
This is on my list to review. I've been looking into more detail on the issue I raised about map tiles not downloading and I think there's some relation, so please stand by a little longer.

oliverlockwood:
@HilaryN Sorry it's taken so long to get back to you on this.  Given the delay, rather than make a bunch of comments, I've instead made one commit on top of your changes here; it's just refactoring and simplifying for maintainability.

I will approve this PR now and give you a chance to review those changes to see if you're happy with them (and maybe give you some learning for future work 😃 ).

Please feel free to merge this yourself, or ask questions, etc.  If I don't hear back by Friday I'll go ahead and merge this myself.

Nice work 🙌

oliverlockwood (approved):

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.kt
```diff
@@ -106,25 +121,30 @@ private fun activityFromContext(initialContext: Context): Activity? {
...
-    if (grantResults.isNotEmpty() && grantResult == PackageManager.PERMISSION_GRANTED) {
+/**
+ * Generic handler for permission Handle
+ *
+ * @param grantResult - either [PERMISSION_GRANTED] or [PERMISSION_DENIED]
+ * @param permission - the string identifier of the permission, e.g. [Manifest.permission.ACCESS_FINE_LOCATION]
+ * @param action - the closure indicating the action to be taken if the permission was granted
+ */
+fun requestPermissionsResultAction(grantResult: Int?, permission: String, action: () -> Unit) {
+    if (grantResult == PERMISSION_GRANTED) {
```
HilaryN, line 147: Just to clarify, grantResults.isNotEmpty() is superfluous if grantResult == PackageManager.PERMISSION_GRANTED ?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.kt
```diff
@@ -31,12 +34,18 @@ fun hasPermission(context: Context?, permission: String): Boolean {
...
 // for some details about the different "requesting permissions" context that Android has.
 
-// Do or request permission - from activity, fragment, or (with hackery) context
-
-fun doOrRequestPermission(activity: Activity?, fragment: Fragment?, permission: String, request_code: Int, action: () -> Unit) {
-    var context: Context? = activity
-    if (fragment != null)
-        context = fragment.requireContext()
+/**
+ * Do or request permission - where you only have a context, so use hackery to derive the activity
```
HilaryN, line 38: This line of comment ("where you only have a context, so use hackery...") doesn't belong here, does it?  It goes with the function below doesn't it?

HilaryN:
Thanks for this and thanks for the learning. :-) It all looks good, with just a couple of small queries above.

I realised from recent reading on Kotlin that (unlike in java) I needn't have gone back and changed all the other calls to doOrRequestPerm to match the parameters - I could have just put in a default value of null for the activity and fragment parms, but never mind, it's done now.

No worries about the delay.  For my first change I was keen to get feedback soon-ish, but now I'm happy to carry on doing other things. (Though I don't want to overwhelm you with too many PR's :-) )

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.kt
oliverlockwood: Every place `requestPermissionsResultAction` is called is within a loop iterating through `grantResults`.  So if `grantResults` were ever empty, this method would not be called.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/util/Permissions.kt
oliverlockwood: Essentially yes, except that this method is the only one that invokes the hackery one 😄 

I've been a bit unhappy with how this works - in the failure case it would throw an NPE through Kotlin's `!!` operator - so I'll push one more commit shortly which will merge these two methods and make the failure case cleaner.

PR #459, opened 2021-03-04
408 circular route 210304

### LINKED ISSUE #408: Feature request: allow circular routes
I saw a review saying this is available on iPhone: allow planning of a circular route from start point, entering number of miles you wish to cycle.

HilaryN:
To fix #408.

I've only done Duration and Distance – not POI's yet.
It works from the map screen but not from the Plan a Route screen where you put in waypoints manually.

Terminology note:  circular and leisure route are the same thing and are used interchangeably, as are linear route and A-B route.

### Summary of changes

- Created activity for input of circular route details (distance or duration).
- Added extra API call for circular route. (Also added error parm to API calls.)
- Amended bits in between front and back end.
- Amended LiveRide replan route to work for circular route.

More detailed notes:
[Circular route notes 210305.pdf](https://github.com/cyclestreets/android/files/6090318/Circular.route.notes.210305.pdf)

Example json for circular/leisure route:
[json leisure.txt](https://github.com/cyclestreets/android/files/6090320/json.leisure.txt)

Test plan/log:
[Circular routes test plan 210305.pdf](https://github.com/cyclestreets/android/files/6090361/Circular.routes.test.plan.210305.pdf)

Scrappy picture of function calls:
Circular route fn calls


libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
```diff
@@ -258,9 +282,9 @@ class TapToRouteOverlay(private val mapView: CycleMapView) : Overlay(), TapListe
     ////////////////////////////////////
     private enum class TapToRoute private constructor(val waypointingInProgress: Boolean, val actionDescription: String) {
         WAITING_FOR_START(false, "Tap map to set Start"),
-        WAITING_FOR_SECOND(true, "Tap map to set Waypoint"),
+        WAITING_FOR_SECOND(true, "Tap map to set Waypoint\nTap here for Circular Route"),
```
oliverlockwood, line 285: I would propose changing this string to: `Tap map to set Waypoint, or\nTap here for Circular Route`

Aside: we should probably move these texts into the languages XML so they can be internationalised

libraries/cyclestreets-view/src/main/res/values/strings.xml
```diff
@@ -33,8 +33,11 @@
   <string name="route_fetching_existing">Fetching route from CycleStreets.net, please wait</string>
   <string name="route_finding_failed">Route finding failed</string>
   <string name="route_loading">Loading route, please wait</string>
+  <string name="route_not_found" translatable="false">Route not found</string>
```
oliverlockwood, line 36: Any reason for making these new strings not translatable?  If it was just "I'm still developing" then fine, but we should change this when all other comments are addressed, and add placeholders (with the English values) in the different languages XML so they're ready to be translated.

libraries/cyclestreets-view/src/main/AndroidManifest.xml
```diff
@@ -13,18 +13,26 @@
...
-    <activity android:name="net.cyclestreets.FeedbackActivity"
-              android:label="Route Comments"
-              android:windowSoftInputMode="stateVisible|adjustResize" />
-    <activity android:name="net.cyclestreets.LiveRideActivity"
-              android:label="LiveRide"
-              android:launchMode="singleTop"
-              android:screenOrientation="portrait" />
+        <activity
+            android:name="net.cyclestreets.views.CircularRouteActivity"
+            android:theme="@style/Theme.AppCompat.DayNight.DarkActionBar">
```
oliverlockwood, line 18: If this _does_ need to be its own activity, why is it themed where others are not? and why is it within `views` package when other activities are not?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/CycleStreetsRoutingTask.kt
```diff
@@ -7,9 +7,12 @@ import net.cyclestreets.view.R
...
                                             private val speed: Int,
-                                            context: Context) : RoutingTask<Waypoints>(R.string.route_finding_new, context) {
+                                            context: Context,
+                                            private val distance: Int? = null,
+                                            private val duration: Int? = null,
+                                            private val pois: String? = null) : RoutingTask<Waypoints>(R.string.route_finding_new, context) {
     override fun doInBackground(vararg waypoints: Waypoints): RouteData? {
         val wp = waypoints[0]
-        return fetchRoute(routeType, speed, wp)
+        return fetchRoute(routeType, -1, speed, wp, distance, duration, pois)
```
oliverlockwood, line 16: I don't like that the interface change you've made forces us to put a magic number -1 here.
If it has to be so, please define a constant something like `NO_ITINTERARY = -1`, and use it both here and within the test in `doFetchRoute()`

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/RoutingTask.kt
```diff
@@ -2,49 +2,84 @@ package net.cyclestreets.routing
...
+                             pois: String? = null): RouteData? {
         return try {
-            val json = doFetchRoute(routeType, itinerary, speed, waypoints)
-            RouteData(json, waypoints, null)
+            val json = doFetchRoute(routeType, itinerary, speed, waypoints, distance, duration, pois)
+
+            when {
+                (json == null) -> {
+                    error = context.getString(R.string.route_not_found)
+                    throw Exception(error)
```
oliverlockwood, line 39: Don't throw raw `Exception`, please 🤮 
If you deliberately want to use a checked exception, then it must be a specific custom subclass, so we can explicitly handle it in a specific way.

As you're just using it to jump to the catch block, I'd accept throwing raw `RuntimeException`, if you don't want to define a custom exception.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/RoutingTask.kt
```diff
@@ -2,49 +2,84 @@ package net.cyclestreets.routing
...
+                }
+                json.contains("Error", ignoreCase = true) -> {
+                    // Put space after ":", remove quotes, drop "{" at start, drop "}" at end
+                    //e.g. {"Error":"The requested itinerary cannot be found."} -> Error: The requested itinerary cannot be found.
+                    error = json.replace(":\"", ": ")
+                                .replace(""""""", "", true)
+                                .drop(1)
```
oliverlockwood, line 46: I absolutely cannot accept manual parsing of a JSON string, and neither should you 😉 

Suggest you instantiate an `ObjectMapper` at appropriate scope (the class?) and then either:
- define a POJO representing the JSON model, and parse into that, or:
- parse into ObjectNode, and then read the value of the `error` field, as explained in https://stackoverflow.com/a/26191723/2108057
`objectMapper.readValue(json, JourneyDomainObject::class.java)

oliverlockwood (changes requested):
Some initial thoughts from a first pass.  I haven't looked at everything but I thought I'd rather give you some feedback now than wait until I've processed it all.

libraries/cyclestreets-view/src/main/res/values/strings.xml
HilaryN: Being lazy, avoiding errors in the build for now.
I'll add them in.
btw Is there an easy way for a translator to see which strings still need translating, other than reading through them all?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/CycleStreetsRoutingTask.kt
HilaryN: I'm pretty sure the -1 was already there in the previous code. (Probably in the FetchCycleStreetsRoutingTask.)  There were 2 fetchRoute functions and one just called the other, so it seemed tidier to combine them.

Happy to make a constant though.

cyclestreets.app/src/main/AndroidManifest.xml
```diff
@@ -88,6 +88,10 @@
         <category android:name="android.intent.category.DEFAULT" />
       </intent-filter>
     </activity>
+    <activity
+        android:name="net.cyclestreets.views.CircularRouteActivity"
```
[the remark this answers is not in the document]
HilaryN: Thanks for these comments.

What would you suggest otherwise, instead of the Activity?  

I originally thought to do an alert dialog, but the number of buttons (including cancel) exceeded the max.  (Bear in mind we still have to add in a button to select POI categories.)

libraries/cyclestreets-view/src/main/AndroidManifest.xml
HilaryN: I wanted to have the more modern theme of the back arrow at the top, which I don't think existed in earlier Android versions.
It wasn't clear to me how the classes are grouped in the packages, so I put it in the same place as the overlays as it's coming from the TapToRouteOverlay - I'm happy to move it to where the other activities are if you prefer.

libraries/cyclestreets-view/src/main/res/values/strings.xml
oliverlockwood: Dunno, Google it and see 🙂

libraries/cyclestreets-view/src/main/AndroidManifest.xml
oliverlockwood: Perhaps you could include some screenshots of the new user journeys etc so I can see what it all looks like without running it locally, and then we can discuss.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/CycleStreetsRoutingTask.kt
oliverlockwood: Indeed, we used an overload method for this. Outside the scope here makes it harder to see what's meant. And anyway, boy/girl scout rule of making things better as you go along, a descriptive constant would indeed be an improvement anyway 🙂

cyclestreets.app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
oliverlockwood: I'll ponder after seeing some screens. You may well be right, I just wanted to discuss

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/RoutingTask.kt
mvl22: Yes, never parse JSON manually is a lesson we learnt :)

oliverlockwood: My rule of thumb is always to ask myself ten times a day "am I doing something tedious which surely must be written better already in a library?"

libraries/cyclestreets-view/src/main/AndroidManifest.xml
HilaryN: Apologies, I should have included a link to the issues thread, where I've described the flow with screenshots:
https://github.com/cyclestreets/android/issues/408

HilaryN: I've added screenshots for the Distance tab and also for LiveRide / Replan route.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
HilaryN: I guess I should do the same for "Tap map to set Waypoint\nTap here to Route" --> "Tap map to set Waypoint, or\nTap here to Route".

Yes, good point, I'll move the strings.  I'm trying to move them whenever I come across them, but I was probably too bogged down in other stuff at the time and then forgot! ( - and forgot to put in a //todo)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/RoutingTask.kt
HilaryN: I must admit, I only covered error handling and exceptions briefly in my reading so far, so this is the perfect cue for me to go and do some more in-depth reading.  Bear with me... :-)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/TapToRouteOverlay.kt
HilaryN: I couldn't work out how to access the context easily from inside the enum class, so have done the context.getString outside the class.  From what I can see, the actionDesc for ALL_DONE doesn't get called, but I've done a check just in case.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/RoutingTask.kt
HilaryN: I've now defined a custom exception but not sure if this was what you had in mind?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/RoutingTask.kt
HilaryN: Using JsonNode seemed the simplest option for a single field.
Having "Error: " in the toast message seems superfluous, so I left that out of the toast message and just display the message itself.
Error message from server as toast

libraries/cyclestreets-view/src/main/res/values/strings.xml
HilaryN: (Still todo: add in all the translation placeholders - will wait til end in case of any changed wording suggestions.)

oliverlockwood:
@HilaryN I'll do a fuller review in a couple of days.  However one thing that occurs to me is you need to update `beta.txt` and `whatsnew.html` to describe the new features.  I've put a line (commented out) into `whatsnew.html` that you may choose to use - if you're happy with it, just comment it out and then add the same text without the HTML markup into the `beta.txt` file.
(I've updated your branch with the latest code from `master`, so you'll need to pull the changes down to your local working copy.)

HilaryN:
> 
> 
> ... - if you're happy with it, just comment it out and then add the same text without the HTML markup into the `beta.txt` file.
> (I've updated your branch with the latest code from `master`, so you'll need to pull the changes down to your local working copy.)

Done this now, but before I push I just want to confirm that I should leave the text commented out in whatsnew?
Or was that a typo and should I _uncomment_ the text in whatsnew?

oliverlockwood:
It was a typo, now corrected 🙂

libraries/cyclestreets-view/src/main/res/layout/activity_circular_route.xml
```diff
@@ -23,11 +25,14 @@
...
         android:layout_marginTop="16dp"
 
+        android:paddingStart="8dp"
         android:paddingTop="8dp"
+        android:paddingEnd="8dp"
         android:paddingBottom="8dp"
         app:tabGravity="fill"
         app:tabIndicatorColor="@color/apptheme_color"
         app:tabIndicatorHeight="4dp"
+        app:tabMaxWidth="0dp"
```
HilaryN, line 35: In Landscape mode the tabs were squashed into the middle of the screen despite having "match_parent on the width". (Sorry I don't have a screenshot of that.)
tabMaxWidth="0dp" allows the tabs to expand to full screen width.

HilaryN:
I've added the ViewModel as previously data entered by the user (Distance/Duration) was lost if the screen was rotated.

oliverlockwood (approved):
@HilaryN as you can see, I've added a number of commits to your changes, largely for maintainability.  The key things were:
- further improving error handling (although you had started using `ObjectMapper` for JSON parsing, you were still string-comparing a JSON string to see if it contained "error", which is bad practice)
- much simplification & clarification of the `CircularRouteActivity` / `CircularRouteViewModel` - enabled by using constants for the tab indexing approach, and separating the logic of "which tab am I on" from "what are the values for each tab", which in turn allowed us to drop the "storeValues" duplication
- tidying up of constants etc.

This is close to being ready to go.  In an ideal world, what I'd like to see is:
- the waypoints that are generated for a circular route being displayed on the map
- translations of the various strings we're adding.

However, as this goes into our beta pipeline, my view (if you're agreeable to it) is to merge this as-is and consider those aspects as improvements to be added subsequently.

On this basis I am approving the PR.  I shall hold off from merging until I hear from you about the above; also, I thought it would be polite to give you a chance to review the changes I've added on top of yours, to check you're happy with them before I merge them in under your name 😄

HilaryN:
Thanks.  Yes, I'm happy with the changes.

Regarding displaying the waypoints, I decided against that as I thought it would unnecessarily clutter the screen and wouldn't add any value.  (See comments in #408 on 21 Feb.)

I guess for consistency they would need to be removed from the itinerary though.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/RetrofitApiClient.java
```diff
@@ -146,13 +148,22 @@ public String getJourneyJson(final String plan,
                                final String leaving,
                                final String arriving,
                                final int speed) throws IOException {
-    Response<String> response = v1Api.getJourneyJson(plan, itineraryPoints, leaving, arriving, speed).execute();
+    Response<String> response = v1Api.getJourneyJson(plan, itineraryPoints, leaving, arriving, speed, REPORT_ERRORS).execute();
```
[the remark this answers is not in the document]
mvl22: > the waypoints that are generated for a circular route being displayed on the map

Just to say that this isn't actually necessary for a circular leisure route. It's arguably perhaps an error that the API actually transmits waypoints other than start/finish waypoint.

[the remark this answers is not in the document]
HilaryN: But they are needed if a replan becomes necessary in LiveRide, if I remember correctly.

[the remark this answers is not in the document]
HilaryN: First 2 screenshots show LiveRide for a circular route.  3rd screenshot shows replan after going off route:
LiveRide replan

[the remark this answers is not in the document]
mvl22: Yes, that's true, so yes, I suppose we do need to transmit them.

However, the point remains that they don't need to be shown in the general route result, as they're really 'virtual' waypoints, since the user never set them.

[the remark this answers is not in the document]
oliverlockwood: OK, fine by me to leave out the waypoints then 👍 

btw I've just had a play with the circular route UI on my phone since the upgrade went out after I merged this, looking good.  Great contribution @HilaryN, thank you again.  I promise to respond promptly once you've had the chance to get the translations prepared!

PR #460, opened 2021-03-22
Gradle upgrades etc.

oliverlockwood:
- move away from JCenter as far as possible
- use `kotlin-stdlib` instead of `kotlin-stdlib-jdk7`
- upgrade `advanced-build-version` plugin
- update UT dependencies all to latest
- upgrade to latest Kotlin version
- upgrade to latest Gradle plugin
- upgrade to latest release of build-tools
- upgrade various Views dependencies


oliverlockwood:
@jezhiggins @mvl22 @HilaryN now Ive got this build reliably green, could I have someone's eyes on to review it, please?

It looks like for whatever reason, without this upgrade, some of the UTs are unreliable on Travis CI.  Which is why Hilary's fix to the circular route API usage has not yet reached the Beta track.

oliverlockwood:
Merging this to unblock the builds. @jezhiggins feel free to comment retrospectively.

jezhiggins:
I think my only comment is that this looked like much harder work than it should have been. Thank you so much for your work on this.

PR #472, opened 2021-04-17
Revert to mikepenz iconics 5.0.3 as icons disappear with 5.2.8

### LINKED ISSUE #471: Live Ride and Plan Route buttons missing
Tapping the place where they used to be still works though, and the text shows if the phone is in landscape mode, so it's just the icon which isn't showing.

HilaryN:
I've not had chance to look into this, so maybe revert to 5.0.3 for now?

Fixes #471


oliverlockwood (approved):

oliverlockwood:
Thanks @HilaryN.  I've raised #473 to track the outstanding work.

PR #482, opened 2021-07-26
Circular Route POIs

### LINKED ISSUE #465: Add POIs option to Circular Route screen
(I've started on this.)

HilaryN:
Fixes #465 

### Brief summary of changes:

- Changed CircularRouteActivity / ViewModel to add in POI button and call to POI list dialog.
- Changed JOLT transform spec to transform POIs (single or multiple) and single or multiple waypoints (rather than just multiple ones as previously).
- Added CircularRoutePoiOverlay as subclass of LiveItemOverlay to display the circular route POIs
- Extracted Bubble functionality from POIOverlay so it can also be used by CircularRoutePoiOverlay
- Added Circular Route POIs to Journey (also stored them in the JourneyDomainObject).

### Other notes:

- In tappedInBubble I've removed the bit which adds a waymark (routeMarkerAtItem) as it doesn't make sense to me to add a waymark?  User can always tap again once the bubble has gone if they want to place a marker.

  @oliverlockwood @jezhiggins Let me know if you disagree, though, and I'll find a way to put it back in for POIOverlay.

- DrawBubble is handled in POIOverlay for both POIOverlay and CircularRoutePOIOverlay.

- I've made Bubble a singleton so that only one bubble will ever be shown, either from the Circular Route POIs or the regular POIs.  (Though, thinking about it, I suppose there wouldn't be much harm in having 2 bubbles – but it's done now.)

- fetchItemsinBackground – the one in CircularRoutePOIOverlay won't get called as onZoom and onScroll are overridden to do nothing.  It's not needed here as the POIs are all uploaded at the beginning for a circular route because there are only a few of them.  Is it OK to just override it to do nothing? (fetchItemsinBackground is called via onZoom and onScroll in LiveItemOverlay).

- Previously I thought that no waypoint was returned for circular routes, but in fact one is returned – it just wasn't getting transformed.  I've corrected the code for this (amended the JOLT spec to transform a single waypoint as well as multiple ones) and removed the code where I previously used the start point (longitude/latitude) instead, in RouteDomainObject and Journey.populateWaypoints.

- Note that the information passed in the API for the circular route POIs is less than that passed for non-route POIs. e.g. journey no. https://cycle.st/j75278584 has bike shops but opening hours aren't shown in the bubble for bike shop “Fully Charged”, whereas they are if you display bike shop POIs from the top right menu. 

- The screen layout design isn't brilliant, sorry!  I was really struggling to get everything in on the Nexus One in landscape mode (without creating a separate layout) and this means a big gap when it's in portrait mode.

[Screenshots of Circular Route with POIs.pdf](https://github.com/cyclestreets/android/files/6878694/Screenshots.of.Circular.Route.with.POIs.pdf)

[Landscape screenshots of Circular Route request screen with POIs button.pdf](https://github.com/cyclestreets/android/files/6878703/Landscape.screenshots.of.Circular.Route.request.screen.with.POIs.button.pdf)

[Test plan for Circular Route POIs.pdf](https://github.com/cyclestreets/android/files/6878700/Test.plan.for.Circular.Route.POIs.pdf)

### DIFF
--- a/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/JourneyPlanner.java
+++ b/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/JourneyPlanner.java
@@ -17,8 +17,8 @@ public class JourneyPlanner {
   public static String getCircularJourneyJson(final Waypoints waypoints,
                                               final Integer distance,
                                               final Integer duration,
-                                              final String pois) {
-    return ApiClient.INSTANCE.getCircularJourneyJson(lonLat(waypoints), distance, duration, pois);
+                                              final String poiTypes) {
+    return ApiClient.INSTANCE.getCircularJourneyJson(lonLat(waypoints), distance, duration, poiTypes);
   }
 
   public static String retrievePreviousJourneyJson(final String plan,
--- a/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/POI.java
+++ b/libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/POI.java
@@ -33,7 +33,7 @@ public class POI
     pos = new GeoPoint(lat, lon);
   }
 
-  void setCategory(final POICategory category) { this.category = category; }
+  public void setCategory(final POICategory category) { this.category = category; }
 
   public int id() { return id; }
   public String name() { return stringOrBlank(name); }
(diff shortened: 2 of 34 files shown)


HilaryN:
PR replaced by #490

PR #483, opened 2021-08-05
Don't save LiveRide replans

### LINKED ISSUE #474: Don't save route replans
If you take a diversion and the app replans your journey, don't save it in the route history. It's annoying especially when you end up with multiple replans.

HilaryN:
Fixes #474 

[Test route saving for LiveRide replans.pdf](https://github.com/cyclestreets/android/files/6938622/Test.route.saving.for.LiveRide.replans.pdf)

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/content/RouteData.java
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/content/RouteData.java
@@ -7,16 +7,20 @@ public class RouteData
   private final String name;
   private final String json;
   private final Waypoints points;
+  private final Boolean saveRoute;
 
   public RouteData(final String json,
                    final Waypoints points,
-                   final String name) {
+                   final String name,
+                   final Boolean saveRoute) {
     this.json = json;
     this.points = points;
     this.name = name;
+    this.saveRoute = saveRoute;
   }
 
   public String name() { return name; }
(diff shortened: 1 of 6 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/content/RouteData.java
```diff
@@ -7,16 +7,20 @@
   private final String name;
   private final String json;
   private final Waypoints points;
+  private final Boolean saveRoute;
```
oliverlockwood, line 10: Any particular reason to make this a `Boolean` object rather than a primitive `boolean`?  Same goes for the constructor param, obviously

oliverlockwood (approved):
One question, but I'm happy with the implementation as it stands, so feel free to ignore if you like and just press the merge button.  Nice and simple - good work!

oliverlockwood:
@HilaryN as noted by email to @mvl22 , the PR build won't run on this for some reason.
I'm going to merge this anyway and let the master build do its business.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/content/RouteData.java
HilaryN: No, no reason - it can be a primitive.

PR #484, opened 2021-08-05
Use current journey plan for LiveRide replans

### LINKED ISSUE #476: Use current route type for replan
When re-planning a route from LiveRide, use the current route type rather than the one in settings.
e.g. I might have "balanced" as preferred type in my settings, but I might decide I want to use the fastest route when riding a particular route.  If I go off route then I'd like the replan to continue to be the fastest route rather than changing it to balanced.

HilaryN:
Fixes #476 

[Test LiveRide replan uses current plan.pdf](https://github.com/cyclestreets/android/files/6940526/Test.LiveRide.replan.uses.current.plan.pdf)

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
@@ -181,6 +181,11 @@ object Route {
         return plannedRoute_
     }
 
+    @JvmStatic
+    fun currentJourneyPlan(): String {
+        return currentJourneyPlan
+    }
+
     private fun loadLastJourney() {
         val routeSummaries = storedRoutes()
         if (!storedRoutes().isEmpty()) {
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
@@ -5,6 +5,7 @@ import android.content.SharedPreferences
 import android.util.Log
 import android.widget.Toast
 import net.cyclestreets.CycleStreetsPreferences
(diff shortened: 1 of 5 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
```diff
@@ -50,11 +52,17 @@ object Route {
...
+    fun LiveReplanRoute(speed: Int,
                         context: Context,
                         waypoints: Waypoints) {
-        val query = LiveRideReplanRoutingTask(plan, speed, context)
+        var newPlan: String
+        // Check current plan is a linear route plan.
+        if (currentJourneyPlan in RoutePlans.allPlans())
+            newPlan = currentJourneyPlan
+        else
+            newPlan = RoutePlans.PLAN_QUIETEST
```
oliverlockwood, line 63: As discussed I think this should be Balanced.

oliverlockwood (changes requested):
I'm happy with this, subject to the suggested change.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Route.kt
HilaryN: I'm happy with Balanced, but it depends on whether Circular routes are consistently planned as quietest or balanced at the back end, in order to stick as closely as possible to the original route.
Hopefully @mvl22 can tell us that.

HilaryN: I made the change, then saw Martin's comment to say Circular routes are based on Quietest, so I've taken the change out again!
I've also changed Boolean to a primitive in RouteData.

oliverlockwood (approved):
Nice work.

PR #490, opened 2021-08-23
465 circular routes pois 210723

### LINKED ISSUE #465: Add POIs option to Circular Route screen

HilaryN:
Fixes #465.

Replaces #482.  Notes and attachments copied over from original PR below:

### Brief summary of changes:

1.  Changed CircularRouteActivity / ViewModel to add in POI button and call to POI list dialog.
2.  Changed JOLT transform spec to transform POIs (single or multiple) and single or multiple waypoints (rather than just multiple ones as previously).
3.  Added CircularRoutePoiOverlay as subclass of LiveItemOverlay to display the circular route POIs
4.  Extracted Bubble functionality from POIOverlay so it can also be used by CircularRoutePoiOverlay
5.  Added Circular Route POIs to Journey (also stored them in the JourneyDomainObject).
 

### Other notes:

7. In tappedInBubble I've removed the bit which adds a waymark (routeMarkerAtItem) as it doesn't make sense to me to add a waymark?  User can always tap again once the bubble has gone if they want to place a marker.

  @oliverlockwood @jezhiggins Let me know if you disagree, though, and I'll find a way to put it back in for POIOverlay.

8. DrawBubble is handled in POIOverlay for both POIOverlay and CircularRoutePOIOverlay.

9. I've made Bubble a singleton so that only one bubble will ever be shown, either from the Circular Route POIs or the regular POIs.  (Though, thinking about it, I suppose there wouldn't be much harm in having 2 bubbles – but it's done now.)

10. fetchItemsinBackground – the one in CircularRoutePOIOverlay won't get called as onZoom and onScroll are overridden to do nothing.  It's not needed here as the POIs are all uploaded at the beginning for a circular route because there are only a few of them.  Is it OK to just override it to do nothing? (fetchItemsinBackground is called via onZoom and onScroll in LiveItemOverlay).
01/12/2021 Edit: Ignore this comment.  I've now removed LiveItemOverlay from CircularRoutePOIOverlay.

11. Previously I thought that no waypoint was returned for circular routes, but in fact one is returned – it just wasn't getting transformed.  I've corrected the code for this (amended the JOLT spec to transform a single waypoint as well as multiple ones) and removed the code where I previously used the start point (longitude/latitude) instead, in RouteDomainObject and Journey.populateWaypoints.

12. Note that the information passed in the API for the circular route POIs is less than that passed for non-route POIs. e.g. journey no. https://cycle.st/j75278584 has bike shops but opening hours aren't shown in the bubble for bike shop “Fully Charged”, whereas they are if you display bike shop POIs from the top right menu. 

13. The screen layout design isn't brilliant, sorry!  I was really struggling to get everything in on the Nexus One in landscape mode (without creating a separate layout) and this means a big gap when it's in portrait mode.

[Screenshots of Circular Route with POIs.pdf](https://github.com/cyclestreets/android/files/7265660/Screenshots.of.Circular.Route.with.POIs.pdf)

[Landscape screenshots of Circular Route request screen with POIs button.pdf](https://github.com/cyclestreets/android/files/7265668/Landscape.screenshots.of.Circular.Route.request.screen.with.POIs.button.pdf)

[Test plan for Circular Route POIs 211001.pdf](https://github.com/cyclestreets/android/files/7271771/Test.plan.for.Circular.Route.POIs.211001.pdf)


oliverlockwood:
@HilaryN could you please resolve the merge conflicts on this PR?

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/JourneyStringTransformer.kt
```diff
@@ -61,6 +77,19 @@ private const val V1API_JSON_JOLT_SPEC = """[{
...
     },
+    "poi": {
+      "*": { "\\@attributes": "pois" }
+    },
+   
     "error": "Error"
   }
-}]"""
+},
+// ... so we need to convert the strings back to arrays!
```
oliverlockwood, line 87: wow this looks like it was tedious 😆

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -151,8 +155,17 @@ class Journey private constructor(wp: Waypoints? = null) {
...
             populateWaypoints(jdo)
             populateSegments(jdo)
+            // Currently (June 2021) the API returns POIs whether or not they were requested.
+            // https://github.com/cyclestreets/android/issues/465#issuecomment-818049555
+            // For simplicity's sake, Will display these, even if not requested,
+            // because, in the event of opening a circular route by number
+            // there is no way of telling whether POIs were requested originally
+            populatePois(jdo, context)
+
+            //
```
oliverlockwood, line 168: I'm not sure this empty comment is necessary

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -200,6 +208,29 @@ class Journey private constructor(wp: Waypoints? = null) {
                 journey.elevation.add(sdo.segmentProfile)
             }
         }
+        // For circular routes.
+        private fun populatePois(jdo: JourneyDomainObject, context: Context) {
```
oliverlockwood, line 212: what stops this running for non-circular routes?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
```diff
@@ -200,6 +208,29 @@ class Journey private constructor(wp: Waypoints? = null) {
...
+            // For circular route pois, the id isn't returned in the API, so will create an artificial one.
+            var id = 0
+            for (poi in jdo.pois) {
+                val circularRoutePoi = POI(id,
+                        poi.name,
+                        "",
```
oliverlockwood, line 218: Don't we have the other fields available for the constructor?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/domain/PoiDomainObject.kt
```diff
@@ -0,0 +1,18 @@
...
+    val name: String? = null
+    @JsonProperty
+    val website: String? = null
+    @JsonProperty
+    val longitude: Float = 0.0F
+    @JsonProperty
+    val latitude: Float = 0.0F
+    @JsonProperty
+    val sequenceId: Short = 0
+}
```
oliverlockwood, line 18: Tedious, but missing newline at end of file (that's what the red line in the circle means in Github)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
```diff
@@ -0,0 +1,122 @@
...
+import androidx.core.content.ContextCompat
+import net.cyclestreets.util.Draw
+import org.osmdroid.api.IProjection
+import org.osmdroid.views.MapView
+import net.cyclestreets.Undoable
+import net.cyclestreets.view.R
+import net.cyclestreets.views.CycleMapView
+import org.osmdroid.views.Projection
+
+object Bubble {
```
oliverlockwood, line 18: For future reference, it would be helpful to reviewers if you add a comment on significant refactors like this to say "This is X extracted from Y" or whatever.  I just figured that out myself.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
```diff
@@ -0,0 +1,122 @@
...
+
+        if (!bubble!!.contains(eventX, eventY))
+            return false
+        // Check if tapped on link
+        if (eventY < Draw.titleSectionY) {
+            showWebpage(activeItem, context)
+        }
+
+        hideBubble(overlay)
+        return true
```
oliverlockwood, line 49: hmmm.  This is different behaviour from the old code in `POIOverlay.kt`, which had:

```kt
       // Check if tapped on link
        if (eventY < Draw.titleSectionY) {
            showWebpage(activeItem)
            return true
        }

        return routeMarkerAtItem(activeItem)
```
Why are we now always hiding  the bubble?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/POIOverlay.kt
```diff
@@ -45,15 +39,15 @@ class POIOverlay(mapView: CycleMapView) : LiveItemOverlay<POIOverlayItem?>(mapVi
...
 
-    private var activeItem: POIOverlayItem? = null
     private var lastFix: IGeoPoint? = null
-    private var bubble: Rect? = null
     private var chooserShowing: Boolean = false
-    private val tapHereText = context.getString(R.string.tap_here)
+
+    init {
+        Bubble.initialise(context, mapView)
+    }
```
oliverlockwood, line 50: Based on `Bubble` being a shared singleton, can you confirm you've done testing which plays around with circular route POIs, and you've then cleared the route and played around with POIs normally, and back again, without issues?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
oliverlockwood: And are we missing `routeMarkerAtItem()` sometimes?

libraries/cyclestreets-view/src/main/res/values/strings.xml
```diff
@@ -183,11 +183,13 @@
...
-  <string name="create_circular_route_button" translatable="false">Create circular route</string>
-  <string name="mins" translatable="false">mins</string>
-
+  <string name="duration">Duration</string>
+  <string name="distance">Distance</string>
+  <string name="circular_route">Circular Route</string>
+  <string name="create_circular_route_desc">Create a circular route based on a duration or distance</string>
+  <string name="create_circular_route_button">Create circular route</string>
+  <string name="mins">mins</string>
+  <string name="choose_pois_to_route_via">Optionally, choose Points of Interest to route via:</string>
```
oliverlockwood, line 192: Would this be better as `Optionally, choose Point of Interest types via which to route:`?

oliverlockwood: (same applies to all untranslated translations, obviously)

oliverlockwood (changes requested):
Looks like a good effort to me.  Some feedback, and some questions.  I'll try to re-review promptly once you've had the chance to address and respond to these!

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
HilaryN: Where's the best place to put that comment?
I mentioned it the summary in the comment on the PR.
Oh wait, sorry, it's in the original PR.  I should have copied over the summary and attachments.

mvl22: As a general point I always treat refactorings of existing code separate to other commits that add new functionality, even if the refactoring is needed because of the addition of that new functionality. This is because the refactoring is a good improvement in itself, and can standalone irrespective of functionality changes.

Using `git add -p` to do partial commits is very helpful for this.

libraries/cyclestreets-view/src/main/res/values/strings.xml
mvl22: While the text in the code isn't perfect queen's english grammar, I think in practice it is clearer.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
oliverlockwood: I tend to agree @mvl22, it's a good practice. Makes things more digestible to review also!

libraries/cyclestreets-view/src/main/res/values/strings.xml
oliverlockwood: Sacrilege! 😂 Fair enough, I'm not hung up on it.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/api/client/JourneyStringTransformer.kt
HilaryN: - and painful... it took ages to work the whole thing out!

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
HilaryN: It will run, but jdo.pois will be empty, so nothing will happen.
Shall I put a condition around the call, or leave it as is in case there is ever a feature introduced to add POIs to a linear route? ;-)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
HilaryN: See note 12 in "Other notes" above.  Only limited info is returned for Circular Route POIs.

HilaryN: The documentation just lists these 6 attributes: longitude, latitude, sequenceId, poitypeId, name, website 
https://www.cyclestreets.net/api/v1/leisure/

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
HilaryN: Yes, it was a lot to plough through!  I was hoping my summary notes would make it a bit more digestible :-)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
HilaryN: Only hiding the bubble if they've tapped on it.  (There is a return a few lines earlier if they tap outside the bubble.)
routeMarkerAtItem also hides the bubble.
See note 7 above as I had a question about routeMarkerAtItem. 
 If we use it, routeMarkerAtItem will only apply to POIOverlay, as CircularRoutePOIOverlay is already showing the route and waymarks can't be added.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/POIOverlay.kt
HilaryN: I always worry I've missed something, so it's good to have your more experienced eye on it.
As I remember, I did extensive testing around the 2 sets of POIs co-existing. (Had some issues which I resolved, but not to do with the Bubble singleton.)
I've added some of these into the test plan attached.  Could you have a look at items 25-38 in the test plan to see if this covers what you had in mind?

HilaryN: There is one slight issue in that where the same POI is displayed as both a normal one and a Circular Route one, the bubble displayed will be the Circular Route bubble, which can have slightly less info than the normal POI bubble.

I tried to reverse this, but it was a bit more work than I originally thought, so I left it for now.  Would it be OK to raise this as a separate issue? (@mvl22, thoughts welcome?)

(I wasn't able to do any clever exclusion of the Circular Route ones from the Normal ones, or vice versa, as I didn't have a unique identifier and it would have been too hit and miss to do it by location, name, etc)

libraries/cyclestreets-view/src/main/res/values/strings.xml
HilaryN: Phew! ;-)  I haven't yet found an easy way to amend all the translations, esp without spaces getting added / removed in the rest of the file and making it look like loads of lines have been amended - I'm sure there must be a way.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Journey.kt
oliverlockwood: fine as-is

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/Bubble.kt
oliverlockwood: Note 7 makes sense.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/POIOverlay.kt
oliverlockwood: seems OK to me.

oliverlockwood (approved):

Issue #486, opened 2021-08-12
Allow insertion / removal of waypoints on Journey Planner screen


HilaryN:
Allow the user to remove waypoints (not just the last one) and insert waypoints before / after another waypoint.

It would be good to have #317 (numbered waypoints) in place first.

PR #491, opened 2021-08-28
Increase waymark size and add waymark number.

### LINKED ISSUE #317: Number the waypoints?
On a Play Store review, a user has suggested "numbered waymarkers to avoid going the wrong way when routes cross."

Something to be looked into, and considered.

HilaryN:
To fix #317 
See what you think about the size, if you want to check it on your phone.  If it's too small, can always increase to 1.8.

### DIFF
--- a/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
+++ b/libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
@@ -29,7 +29,7 @@ class WaymarkOverlay(private val mapView: CycleMapView) : Overlay(), PauseResume
     private val screenPos = Point()
     private val bitmapTransform = Matrix()
     private val bitmapPaint = Paint()
-    private val boldTextBrush = Brush.createBoldTextBrush((offset(mapView.getContext())*0.8).toInt())
+    private val waymarkNumberTextBrush = Brush.createBoldTextBrush((offset(mapView.getContext())*0.8).toInt())
 
     private val waymarkers = ArrayList<OverlayItem>()
 
@@ -106,7 +106,7 @@ class WaymarkOverlay(private val mapView: CycleMapView) : Overlay(), PauseResume
                            index: Int,
                            size: Int) {
 
-        val waymarkPosition = when (index) {
+        val waymarkNumber = when (index) {
                                     0 -> "S"                    // Starting waymark
                                     size - 1 -> "F"             // Finishing waymark
                                     else -> {index.toString()}  // Numbered intermediate waymark
@@ -136,11 +136,15 @@ class WaymarkOverlay(private val mapView: CycleMapView) : Overlay(), PauseResume
 
         val x = screenPos.x.toFloat()
         val y = screenPos.y.toFloat()
+        // Coordinates for Waymark number (position it within body of wisp):
+        val wmNumXCoord = x - halfWidth/10
+        val wmNumYCoord = (y - halfHeight/1.8).toFloat()
+
         canvas.apply {
             save()
             rotate(-projection.orientation, x, y)
             drawBitmap(bitmap, bitmapTransform, bitmapPaint)
-            drawText(waymarkPosition, x - halfWidth/10, (y - halfHeight/1.8).toFloat(), boldTextBrush)
+            drawText(waymarkNumber, wmNumXCoord, wmNumYCoord, waymarkNumberTextBrush)
             restore()
         }
     }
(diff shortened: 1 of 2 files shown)


libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -120,10 +134,13 @@ class WaymarkOverlay(private val mapView: CycleMapView) : Overlay(), PauseResume
...
         }
 
+        val x = screenPos.x.toFloat()
+        val y = screenPos.y.toFloat()
         canvas.apply {
             save()
-            rotate(-projection.orientation, screenPos.x.toFloat(), screenPos.y.toFloat())
+            rotate(-projection.orientation, x, y)
             drawBitmap(bitmap, bitmapTransform, bitmapPaint)
+            drawText(waymarkPosition, x - halfWidth/10, (y - halfHeight/1.8).toFloat(), boldTextBrush)
```
oliverlockwood, line 143: Magic number 1.8 here.  Should this be `INCREASE_WAYMARK_SIZE` (i.e. currently 1.5) or a new constant?
Also what's with `halfWidth/10`?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
```diff
@@ -30,6 +29,7 @@ class WaymarkOverlay(private val mapView: CycleMapView) : Overlay(), PauseResume
     private val screenPos = Point()
     private val bitmapTransform = Matrix()
     private val bitmapPaint = Paint()
+    private val boldTextBrush = Brush.createBoldTextBrush((offset(mapView.getContext())*0.8).toInt())
```
oliverlockwood, line 32: another magic number `0.8` here - what does this mean?

oliverlockwood (commented):
A couple of questions

oliverlockwood:
Given that the wisps are bitmaps create from the vectors, it would be better to recreate the bitmaps at the increased (x 1.5) size, or move to directly using the vectors as noted in comments on the issue.

Scaling a bitmap results in blurry images, reducing the visual quality of the app.

I'm OK with going with this solution as an interim only if another ticket is raised, highlighting where the scaling currently is such that it can be ripped out when the bitmaps are updated.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
HilaryN: IIRC both of those numbers are to position the text within the waymark.  Let me try and think of appropriate constant names... 
HORIZONTAL_TEXT_POSITION_ADJUSTMENT, VERTICAL_TEXT_POSITION_ADJUSTMENT or something like that.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
HilaryN: Text size.

HilaryN:
Ah, as it happens I'm starting to look at both those options for the next piece of work on #486 as I want to use ItemizedOverlay for drawing the waymarks so that I can then use the HitTest to see if a waymark has been tapped on and make it the active item etc.  (Currently the waymark is in a massive square and I'd like to crop it, otherwise it will appear they have tapped on the item when tapping a point a long way away.)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
HilaryN: Those constants are a bit of a mouth / eye-ful, so I've created more descriptive variables instead.

HilaryN:
I've renamed WaymarkPosition to WaymarkNumber.  I know it's not totally accurate but it's a better description, as position might imply position on screen.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/WaymarkOverlay.kt
oliverlockwood: Also I would extract all constants to top level for the class - not inline in specific functions/methods as you have currently.

Am ready to re-review your update when it is good to go - ping me at that time please.

HilaryN: OK, done, and new issue raised for the bitmaps - #498

oliverlockwood (approved):

PR #499, opened 2021-12-10
498 recreate waymark bitmaps

### LINKED ISSUE #498: Recreate waymark bitmaps
Recreate the waymark bitmaps at the increased (x 1.5) size, or move to directly using the vectors (see comments in #317).

HilaryN:
Fixes #498.

As noted in PR #365 the outline disappeared when converting wisp.svg to a vector drawable.  There's a fix proposed here:
https://stackoverflow.com/questions/63655388/android-vector-why-is-the-stroke-color-not-adopted
but I didn't like the idea of repeating the path data, so I did a fudge of adding in a dummy gradient for the outline stroke and that worked.  Maybe there's a better way but I gave up...

I also moved the waymark drawing to ItemizedOverlay to simplify it and because I'll need this for the next piece of work to see if user has tapped on a waymark.

Screenshot_1639172025

Screenshot_1639172052


oliverlockwood:
@HilaryN - should I review this first in isolation and then you'll update your other PR to account for it? Or what's the logical approach?

HilaryN:
Probably best to finish off the other PR as this is built on top of that one.

oliverlockwood:
@HilaryN now I've merged #491, perhaps you can update this to resolve conflicts etc.

Please make sure in your testing that placing waymarks puts the point of the wisp onto where you tap the screen (when I created my custom sized bitmaps, you will notice there was transparent space below the wisp to ensure that the wisp image, centred on the tap location, pointed at the tap location).

HilaryN:
Thanks - will do.

Yes, I read about the issue you resolved where the wisp wasn't appearing at the correct place.
I can confirm that the bottom pointy bit of the wisp appears at the tap location (easier to see when testing in the emulator as I can see exactly where I tap with the mouse more easily than tapping with a finger).
I know your image was in a big rectangle whereas mine is in a rectangle exactly sized to the wisp (which in turn makes it easier to see if the waypoint is tapped for my next enhancement of removing / renumbering waypoints).

HilaryN:
Oops!  I must not have tested this on the Nexus.  I've realised that the waymarks are massive on the Nexus 6 and quite small on the Nexus One, meaning that the waymark number is in the wrong place on these phones.

Waymarks and POIs

I found on debugging that the icons are being scaled both at the point where the Drawable is created (e.g. makeWisp / getDrawable for waymarks, PoiTypesDto.poiIcon) and again when they are drawn on the screen, resulting in inconsistent sizes across phones – not a big issue in itself, but it makes it difficult to place the text of the waymark number, which is only being scaled once and so doesn't match the image size.

The same inconsistency of size is occurring for the photomap images and POI icons:

Photomap

I'm not going to change the photomap icons or the POI icons, but I've made a change to stop the waymarks being scaled at the point of being drawn so that the waymark number can be positioned correctly:

Waymarks POIs after

libraries/cyclestreets-view/src/main/java/net/cyclestreets/views/overlay/ItemizedOverlay.java
```diff
@@ -40,19 +46,23 @@ public void draw(final Canvas canvas, final MapView mapView, final boolean shado
...
     }
   }
 
+  float scale(final MapView mapView) {
+    return mapView.getContext().getResources().getDisplayMetrics().density;
+  }
+
   private void onDrawItem(final Canvas canvas,
                           final Item item,
+                          final int index,
```
oliverlockwood, line 65: Please could we rename this parameter to `itemIndex`...
- here
- in the declaration of the `drawTextOnMarker` in this class
- similarly, in the declaration of the overrides `drawTextOnMarker` method in `WaymarkOverlay`

oliverlockwood (approved):
Addressed my own (very minor) comments in a commit.  I'm happy for this to go in once the build is green.

PR #500, opened 2021-12-17
patch(data loss) 1

Naitim:
Hello developers of CycleStreets

I am using your app CycleStreets. I think the app is great but I have one minor patch that could improve the user experience.

Here are some pictures to help illustrate my patch: https://imgur.com/a/S7iCRWB

In picture 1, a user has partially filed out his registration information. If something like a phone call moves the screen to another activity or if the user accidentally closes the app or presses back. The app should preserve the data in this case. However, once the user returns he will lose information he had filed in so far. This can be seen in pictures 1 and 2. Our patch will preserve this data so there will be no data loss in situations like this.

Another issue is when the user finishes registration. When the user wants to register another account. The information of the previous account is pre-filled out as can be seen in picture 3. This is inconvenient as they have to manually delete the data. It could also be misleading since typically, in other apps, you would preserve this data when registration is not complete. Our patch will clear this data if registration is successful.

I have tested out the feature to ensure it works.

If you have any questions or if you would like me to change anything, please do not hesitate to let me know!

Thank you for your time,
Tim

### DIFF
--- a/libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
+++ b/libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
@@ -295,7 +295,6 @@ public class CycleStreetsPreferences
     editor.putString(PREF_CONFIRM_PASSWORD_KEY, confirmPassword);
     editor.putString(PREF_NAME_KEY, name);
     editor.putString(PREF_EMAIL_KEY, email);
-    editor.putBoolean(PREF_VALIDATED_KEY, false);
     editor.commit();
   }
 
@@ -308,6 +307,16 @@ public class CycleStreetsPreferences
     editor.putString(PREF_CONFIRM_PASSWORD_KEY, "");
     editor.putBoolean(PREF_PENDING_KEY, false);
     editor.putBoolean(PREF_VALIDATED_KEY, false);
+    editor.commit();
+  }
+
+  public static void clearTempUsernamePassword() {
+    final Editor editor = editor();
+    editor.putString(PREF_USERNAME_KEY, "");
+    editor.putString(PREF_PASSWORD_KEY, "");
+    editor.putString(PREF_NAME_KEY, "");
+    editor.putString(PREF_EMAIL_KEY, "");
+    editor.putString(PREF_CONFIRM_PASSWORD_KEY, "");
     editor.commit();
   }
 
--- a/libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
+++ b/libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
@@ -285,18 +285,16 @@ public class CycleStreetsPreferences
   }
 
   public static void setTempUsernamePassword(final String username,
-                                             final String password,
-                                             final String confirmPassword,
-                                             final String name,
-                                             final String email,
-                                             final boolean pending) {
+                                                final String password,
+                                                final String confirmPassword,
+                                                final String name,
+                                                final String email) {
     final Editor editor = editor();
     editor.putString(PREF_USERNAME_KEY, username);
     editor.putString(PREF_PASSWORD_KEY, password);
     editor.putString(PREF_CONFIRM_PASSWORD_KEY, confirmPassword);
     editor.putString(PREF_NAME_KEY, name);
     editor.putString(PREF_EMAIL_KEY, email);
-    editor.putBoolean(PREF_PENDING_KEY, pending);
     editor.putBoolean(PREF_VALIDATED_KEY, false);
     editor.commit();
   }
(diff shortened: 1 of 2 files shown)


mvl22:
@HilaryN @jezhiggins @oliverlockwood Could one of you review?

HilaryN:
Great to have your contribution, Tim.
It looks good to me - I'll have a more detailed look at the weekend.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
```diff
@@ -283,6 +284,23 @@ public static void setPendingUsernamePassword(final String username,
...
+                                             final String name,
+                                             final String email,
+                                             final boolean pending) {
+    final Editor editor = editor();
+    editor.putString(PREF_USERNAME_KEY, username);
+    editor.putString(PREF_PASSWORD_KEY, password);
+    editor.putString(PREF_CONFIRM_PASSWORD_KEY, confirmPassword);
+    editor.putString(PREF_NAME_KEY, name);
+    editor.putString(PREF_EMAIL_KEY, email);
+    editor.putBoolean(PREF_PENDING_KEY, pending);
```
HilaryN, line 299: I'm thinking it's better to leave PREF_PENDING_KEY out.  Setting it to true displays message saying that they have already tried  to register an account, which they may not have done yet.

libraries/cyclestreets-core/src/main/java/net/cyclestreets/CycleStreetsPreferences.java
```diff
@@ -283,6 +284,23 @@ public static void setPendingUsernamePassword(final String username,
...
+    editor.putBoolean(PREF_VALIDATED_KEY, false);
+    editor.commit();
+  }
+
   public static void clearUsernamePassword() {
     final Editor editor = editor();
```
HilaryN, line 305: Could also clear PREF_CONFIRM_PASSWORD_KEY in clearUsernamePassword()

libraries/cyclestreets-view/src/main/java/net/cyclestreets/AccountDetailsActivity.java
```diff
@@ -77,8 +79,38 @@ public void onCreate(final Bundle saved) {
     setupView();
   }
 
+  @Override
+  protected void onPause() {
+    super.onPause();
```
HilaryN, line 84: What do you think about adding something like the following comment in onPause and onBackPressed:
"If successfully registered, clear user details."
Maybe put that common bit of code into a separate method along with the comment, to avoid repetition.

HilaryN (commented):
Thanks, Tim - I've added some comments :-)

libraries/cyclestreets-view/src/main/java/net/cyclestreets/AccountDetailsActivity.java
mvl22: If there is a success response from the API, then we know it can be cleared. Definitely avoid putting that work on the user.

HilaryN: Are you referring to the PREF_PENDING_KEY comment above?

Naitim:
Hey, thanks for the response. So I will remove the save to PREF_PENDING_KEY and add functionality to clear PREF_CONFIRM_PASSWORD_KEY. Is there anything else I should change?

libraries/cyclestreets-view/src/main/java/net/cyclestreets/AccountDetailsActivity.java
HilaryN: Hi Tim.

Plus the 3rd comment I made:

> What do you think about adding something like the following comment in onPause and onBackPressed: "If successfully registered, clear user details." Maybe put that common bit of code into a separate method along with the comment, to avoid repetition.

HilaryN:
Hi Tim.

Plus the 3rd comment I made above:

> What do you think about adding something like the following comment in onPause and onBackPressed: "If successfully registered, clear user details." Maybe put that common bit of code into a separate method along with the comment, to avoid repetition.

Naitim:
Hey, I've pushed a new commit for you to review with all the recomendded changes. If I missed anything or if you would like me to change anything, let me know.

HilaryN:
This still isn't quite right.  If I submit an account to be registered, then go back in to register an account, it no longer says an account registration is pending.
Btw I'm wondering, under what scenario would a user be creating more than one account?

Naitim:
I believe the issue was because clearUsernamePassword method was setting pending to false so I created a new method(clearTempUsernamePassword) that does not change the pending preference. 

The primary reason for this patch is to prevent data loss during the account creation step. If the user is creating a new account and for some reason(such as a phone call) the user leaves the app and it is force closed(whether by the user or by the android system to clear up RAM), then this patch will allow cyclestreets to automatically restore the data the user has already submitted when they return to the account registration activity. This will improve the user experience whether it is the first account the user is making or if it is another account.

Although to answer your question, perhaps if the user wanted to use a different username or email address, they might want to create a new account in that situation.

HilaryN (approved):
Looks good - thanks again, Tim.

HilaryN:
Fixes #501

HilaryN:
@Naitim do you fancy tackling any other changes?  Your contribution would be very welcome if you do.  :-)
Loads of issues in the list to choose from - give me a shout if you've any questions or want any suggestions.

PR #503, opened 2022-02-15
487 create new route from existing route

### LINKED ISSUE #487: Create new route from existing route
Have a facility to create a new route using an existing route as the basis.

This would copy and display the waypoints from the existing route onto the JourneyPlanner screen.

#486 would ideally be implemented first.

This would make #485 redundant.

HilaryN:
Fixes #487 

On long tapping a displayed route, there is an extra menu option to Add / Remove waypoints.  Taking this option takes you back to the route planner screen with the waypoints from the current route re-displayed.

I haven't put in a confirmation when “Add / Remove waypoints” selected, because if it's done by accident it's not a big deal.  The waypoints are still there and you can just tap to route to get the route back, so nothing is really lost.

The option to Add / Remove waypoints isn't available on Circular routes. (Though I guess it's possible someone might want to copy the starting point to re-do the circular route with different options, e.g. different length, different POI's.)

Screenshot of new menu option to create new route from existing waypoints

[Test plan - 487 create new route from existing.pdf](https://github.com/cyclestreets/android/files/8073228/Test.plan.-.487.create.new.route.from.existing.pdf)


libraries/cyclestreets-view/src/main/res/values-es/strings.xml
```diff
@@ -34,12 +34,13 @@
...
   <string name="route_menu_change_share">Compartir ruta</string>
   <string name="route_menu_change_comment">Comentar ruta</string>
-  <string name="route_menu_change_replan_fastest">Replantear como ruta rápida</string>
-  <string name="route_menu_change_replan_balanced">Replantear como ruta equilibrada</string>
-  <string name="route_menu_change_replan_quietest">Replantear como ruta tranquila</string>
-  <string name="route_menu_change_replan_shortest">Replantear como ruta corta</string>
+  <string name="route_menu_change_replan_fastest">Replanear como ruta rápida</string>
+  <string name="route_menu_change_replan_balanced">Replanear como ruta equilibrada</string>
+  <string name="route_menu_change_replan_quietest">Replanear como ruta tranquila</string>
+  <string name="route_menu_change_replan_shortest">Replanear como ruta corta</string>
```
oliverlockwood, line 43: your knowledge of Spanish (and French) is obviously broader than mine :)

HilaryN: :-)

PR #505, opened 2022-04-25
485 allow addition of extra waypoints to route 220216

### LINKED ISSUE #485: Allow addition of extra waypoint(s) to planned route
Allow user to add extra waypoint(s) to an already-planned route.  Use cases:

1) Have planned a route with a few waypoints and realised you need to add another one.  Don't want to have to tap in all the waypoints again.
2) The route goes through a park and you want to avoid the park as it is night-time.  
3) There is a road closure you know of and you need to avoid it.
4) Tower Bridge is stuck open ;-)

This could be instigated via a menu option and would apply to linear routes only, not circular routes.
When the user taps in an additional waypoint they could be then given 2 options:

1) For the app to automatically calculate the sequencing of the waypoint in relation to the existing ones (if that's possible?)
2) To be presented with a list of already-existing waypoints on the route and select the one after which they want the newly-added one to go.  
(If there is only a start and finish waypoint this step could be skipped.)

The app would then fetch the new route with the extra waypoint.

HilaryN:
Fixes #485 

When a route is displayed, user can tap on the screen to add a waypoint.  The app will determine the waypoint's sequence and then go to fetch the alternative route and display it in a different colour alongside the current route.  See attachment for screenshots:

Screenshots of alt route

The calculation of the waypoint's sequence isn't perfect, of course, and could probably be improved, but it works in straightforward cases, where the existing waypoints are in a logical sequence.  It finds the closest existing waypoint and checks the waypoints before and after this point to decide whether to place it before or after the closest one.

I'm not blocking user entry while the alternative route is being retrieved, so the user is free to tap another waypoint on the screen and then the same thing will happen.  It will cancel the previous task in this case so if the user taps, for example, 6 waypoints in quick succession, only the last will retrieve the alternative route (unless the retrieval is quicker than the user tapping waypoints of course).

I'm assuming that:

“Starting with [Build.VERSION_CODES.HONEYCOMB](https://developer.android.com/reference/android/os/Build.VERSION_CODES#HONEYCOMB), tasks are executed on a single thread to avoid common application errors caused by parallel execution. “
https://developer.android.com/reference/android/os/AsyncTask

Assumed that Start and Finish will remain the same, no matter where additional waypoints are added.  User can always use the “Add / Remove waypoints” option if this isn't the case.

Each time an alt waypoint is added or removed, the alt route will be re-calculated (first cancelling previous task if there is one).

On pause / resume, alt waypoints and route will be re-displayed, but if app is destroyed, alt route will be lost.

The routingInfoRect was used as both a button to plot route and as a text summary of the route.  I've split this out and created a separate button and text view as I now need both simultaneously.

stepBack fun – this was too multi-purpose and too much going on here.  It was called both to start a new journey and to go back a step (back button pressed or a waypoint removed).
I've split it out and created a separate function startNewJourney as there wasn't much functionality in common.

[485 Test plan.pdf](https://github.com/cyclestreets/android/files/8557746/485.Test.plan.pdf)
.


jezhiggins (approved):

jezhiggins:
I feel like there's a more pleasing way we can handle showing more than one route without the `if(alt)` guards, but I think we're better to merge this now and refine it later.

HilaryN:
Thanks for reviewing that.

Yes - I was thinking to make a new class for AltRouteOverlay as an extension of RouteOverlay.  Is that what you have in mind?

It was on my to-do list, but fell off due to lack of time!

PR #517, opened 2022-09-01
497 segment showing as walking

### LINKED ISSUE #497: Cycle segment showing as walking
e.g. in the following journey:
https://cycle.st/j78841326 - Lennard Road to Alley from Park Road
Park Road is showing as a walking segment when it should be cycling.
On the website it shows a a cycling segment.
It may be something to do with 2 (short?) segments being combined.

### LINKED ISSUE #515: Footprints in Itinerary don't show
The footprints, which indicate that a segment is a walking segment in the Itinerary, don't show unless a segment is selected, because they are white on a white background.  It would be nice to be able to see these even if a segment isn't selected.

image

HilaryN:
To fix #497 and partially fix #515 

These images show an example of a journey which has a cycling-then-walking (turn right then turn left) segment pair, followed by a walking-then-cycling (turn left then turn right) segment pair.
The segments are kept as 2 segments, but the first of the pair has the combined instruction (e.g. turn right then turn left) and the street of the second segment of the pair (e.g. "Crossing").  The second segment of each pair just has its original instruction.

Website and app itinerary 84251490

Website and app map 84251490

If the 2 segments are both cycling or both walking then they are melded into one segment as previously.

I've done the same for Bridges (see #497 for fudged example).

I've added tests (SegmentTest) for cycling/walking pairs.  The original tests for melding into one segment are still in LiveRideVoiceTest.

I attempted to fix #515 by making the footprints black, but they still don't show, e.g. on left turn.  I think a separate ImageView will be needed but I'll leave that for now.
In this example the footprints show:
image

but in this example they don't show on the left turn unless that instruction is selected:

image


HilaryN:
See #515 for screenshot with new footprints image.

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Segments.kt
```diff
@@ -44,16 +44,56 @@ class Segments : Iterable<Segment> {
...
+                    if (previous.walk == seg.walk) {
+                        // Combine the 2 segments:
+                        segments.add(Segment.Step(previous, seg, it, it.textInstruction))
+                        return
+                    }
+                    else {
+                    // If the 2 segments aren't both walking or both cycling, keep the segments separate.
+                    // Add previous segment with combined turn instruction and CURRENT street,
+                    // so user still gets warning of the double turn:
+                        segments.add(Segment.Step(
```
oliverlockwood, line 56: Massively redundant block here.  Just move `segments.remove()` into the `if` branch and then you don't need an `else` branch, and all its noisy comments, at all.

libraries/cyclestreets-view/src/test/java/net/cyclestreets/routing/SegmentsTest.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private fun loadJourneyFrom(domainJsonFile: String) {
+        val rawJson = TestUtils.fromResourceFile(domainJsonFile)
+        val routeData = RouteData(rawJson, null, "test route", false)
+        Route.onNewJourney(routeData)
+        journey = Route.journey()
+    }
+
+}
```
oliverlockwood, line 63: Github is complaining about a missing newline at end of file.  It's trivial, but please oblige it.

libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/itinerary/SegmentAdapter.kt
```diff
@@ -18,7 +18,7 @@ import net.cyclestreets.util.Turn
 import net.cyclestreets.util.TurnIcons
 
 internal class SegmentAdapter(context: Context) : BaseAdapter() {
-    private val footprints: Drawable = ResourcesCompat.getDrawable(context.resources, R.drawable.footprints, null)!!
+    private val footprints: Drawable = ResourcesCompat.getDrawable(context.resources, R.drawable.footprints2, null)!!
```
oliverlockwood, line 21: We need to talk about images (see https://github.com/cyclestreets/android/issues/515#issuecomment-1236974153).

libraries/cyclestreets-view/src/main/java/net/cyclestreets/routing/Segments.kt
HilaryN: In the else branch, I'm effectively modifying the previous segment to change the previous instruction and street, so the user still gets warning of the double turn.
If I put the remove inside the if branch and take out the else branch, the user won't get warning of the double turn.  I thought it would be nice to still have the double turn instruction, even for mixed cycling / walking (esp where the first segment of the 2 is a cycling one).

libraries/cyclestreets-view/src/test/java/net/cyclestreets/routing/SegmentsTest.kt
HilaryN: Done.

libraries/cyclestreets-view/src/test/resources/journey-rightleft-walkcycle-domain.json
```diff
@@ -0,0 +1,207 @@
...
+      "startBearing": "17",
+      "color": "#ff0000",
+      "points": "-0.03292,51.42814 -0.03291,51.42816 -0.03282,51.42843 -0.03278,51.42851 -0.03267,51.42857 -0.03254,51.42864 -0.03248,51.42873 -0.03242,51.42889 -0.03229,51.42904 -0.03216,51.42919 -0.03206,51.42945 -0.032,51.42967 -0.03195,51.42976 -0.03185,51.42984 -0.03171,51.42998 -0.03161,51.43008 -0.03153,51.4302",
+      "distances": "0,2,31,9,10,12,11,18,19,19,30,25,11,11,18,13,15",
+      "elevations": "24,24,24,24,23,22,22,21,22,24,25,25,24,24,24,25,25",
+      "provisionName": "Cycle path",
+      "type": "segment"
+    }
+  ]
+}
```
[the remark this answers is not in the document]
HilaryN: Done.

libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/itinerary/SegmentAdapter.kt
HilaryN: Might be better to keep this as a separate issue.

PR #529, opened 2023-08-03
Bump java version

jezhiggins:
Move to Target 31 finally built green, but then the publish step fell over. 

So, we've got to bump to Java 11, and that's triggered it's own series of cascading changes :roll_eyes:


jezhiggins:
OK, let's see if this will make it all the way to publishing :crossed_fingers:

PR #531, opened 2023-08-03
Set JAVA_HOME to 8

jezhiggins:
Maybe this'll work. Who knows?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ before_install:
 install:
   - ls -l /usr/lib/jvm/
   - ./checkPoiTypeIcons.sh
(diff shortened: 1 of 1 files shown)


jezhiggins:
Nope.

oliverlockwood:
If building in Travis is being a pain, you could look into what the Android OSS community now uses for free builds - maybe Github Actions or similar is a better option now?

jezhiggins:
Worked it out in the end - needed to make sure the publish step was using the same JDK as the build. 

I may look at moving to Actions in due course, but since I didn't realise until the end it was a build problem, rather than a plugin/dependency/deprecation/anything else problem, I was trying to make the smallest possible set of changes :)

PR #535, opened 2023-08-20
Target API 33

jezhiggins:
Having just moved to SDK level 31, we're already being pushed to target 33. 

Might as well try and update everything else - gradle, kotlin, dependencies, etc - while we're at it and get everything done in one fell swoop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,15 +20,15 @@ language: android
 addons:
   apt:
     packages:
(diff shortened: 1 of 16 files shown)


jezhiggins:
Initial pass through looks ok, everything seems to be running as it should.

oliverlockwood (approved):
LGTM

PR #543, opened 2023-12-29
542 write external storage deprecated

### LINKED ISSUE #542: Can't add photos to photomap any more
A user reports:

I tried to add a picture of cycle parking to the photomap today, and wasn't able to! It said that I'd denied permission to access my device storage or camera, when I hadn't. I looked in my app settings, and it didn't look like I could grant the app permission for this either. I'm using a Google Pixel 5.

I tried force-quitting to see if the permission would be requested, and even tried uninstalling and reinstalling the app and had the same issue. Looks like it doesn't recognise that it needs storage/photo permission.

Screenshot_20231125-191627
Screenshot_20231125-191704

HilaryN:
To fix #542 

I haven't touched the READ permission request in MapPack as I think that isn't working at the moment?  

Write permissions for app-specific storage (i.e. map tiles) is no longer required.  Same for Read permission for app-specific storage (which doesn't seem to get explicitly requested, presumably as it's granted as part of the write permission).  Removed for Android 13 upwards. (CycleMapFragment)
Note that because permission for map tiles is no longer granted, the old cache location won't get removed – I'm not sure if this is an issue?  (See CycleMapFragment.onRequestPermissionsResult)

CycleMapView – I've just changed the log message in here.

Write permission to store photos should be handled by the camera app – removed for Android 13 upwards. (AddPhotoFragment)

Read permission to access photos - changed to READ_MEDIA_IMAGES for Android 13 upwards.

Tested on Pixel 3 API 28 (Android 9) emulator, Fairphone 2 (Android 10), Fairphone 3 (Android 13).

### DIFF
--- a/cyclestreets.app/src/main/play/release-notes/en-GB/beta.txt
+++ b/cyclestreets.app/src/main/play/release-notes/en-GB/beta.txt
@@ -1,0 +1,2 @@
+- If there are queued voice instructions, these are replaced with the latest instruction.
+- Update permissions for Android 13 so map tiles can be cached and photos added.
--- a/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleMapFragment.kt
+++ b/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleMapFragment.kt
@@ -2,9 +2,8 @@ package net.cyclestreets
 
 import android.Manifest.permission.WRITE_EXTERNAL_STORAGE
 import android.graphics.drawable.Drawable
+import android.os.Build
 import android.os.Bundle
-import androidx.fragment.app.Fragment
-import androidx.preference.PreferenceManager
 import android.util.Log
 import android.view.LayoutInflater
 import android.view.Menu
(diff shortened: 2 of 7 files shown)


mvl22:
Thanks for looking at this, Hilary. I'll leave others to review but it's great to see this being addressed.

mvl22:
@jezhiggins @oliverlockwood Could one of you kindly review this? Seems this will also fix #544.

jezhiggins (approved):

mvl22:
Thanks for the review :)

Are we able to roll a release? Another report about missing map tiles recently.

jezhiggins:
Now this has been merged, the beta release should go out automatically and chances are people will have it by the morning. Assuming that looks ok, we can promote it to a regular release in the next few days.

mvl22:
Awesome, thanks all. Will look forward to the release notes appearing.

HilaryN:
I don't think this has made it to the Play Store yet.  I tried re-installing from there but the issue is still occurring.

jezhiggins:
Yes, I noticed this myself this morning. Going to have a look around this evening or tomorrow. (It's entirely possible there's something I need to do but have forgotten) On 3 Feb 2024 09:30, HilaryN ***@***.***> wrote:
I don't think this has made it to the Play Store yet.  I tried re-installing from there but the issue is still occurring.

jezhiggins:
Something up with Travis - it's stopped building because some kind of problem with our plan. We're on their open source software plan so not sure what's going on, hopefully just a cock up. I've emailed support, they were very quick to respond last time so fingers crossed.

oliverlockwood:
FYI, this happened to the OSS repos we (mostly I) maintain at my employer, too. Travis seems to be gradually making more and more hoops for OSS to jump through, it seems as though in their push to become profitable they're trying to discourage OSS usage while still claiming they support it for free.

I've migrated those to Github Actions, it wasn't too tough and I have a reference point there (though admittedly not for an Android build). I'll take an action to look into doing likewise for the CycleStreets CI at some point.

oliverlockwood:
See e.g. (partly for my own reference):
- https://github.com/android-actions/setup-android
- https://github.com/marketplace/actions/automated-build-android-app-with-github-action

jezhiggins:
Travis have done whatever the need to do, and the build is running again for now. Beta has been pushed up to Google Play so hopefully will make it's way out this evening

PR #549, opened 2024-03-14
Resetting the OSMDroid tile cache.

jezhiggins:
This is relatively drastic, but we've got a number of reports where the app stops drawing tiles when the phone OS has been upgraded. We don't currently detect that situation (which can require a cahce reset), but if we try this on our beta channel we can at least get an idea if it'll help.

If yes, we can do it properly. Otherwise we can roll back.

### DIFF
--- a/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleStreetsAppSupport.java
+++ b/libraries/cyclestreets-fragments/src/main/java/net/cyclestreets/CycleStreetsAppSupport.java
@@ -10,9 +10,7 @@ import android.content.SharedPreferences;
 import android.util.Log;
 
 public final class CycleStreetsAppSupport {
-
   private static final String TAG = Logging.getTag(CycleStreetsAppSupport.class);
-
   private static boolean isFirstRun;
   private static boolean isNew;
   private static String version;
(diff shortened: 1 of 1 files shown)


mvl22:
> which can require a cache reset

Is that a standard Android requirement, documented somewhere?

jezhiggins:
No, it's not an Android requirement. It's to do with how Osmdroid sets up its cache on first use. When the Android version changes its possible for the stored cache location to become unavailable - because mount points have changed, or because of different permission regimes, (and potentially other things in not aware of).There's an element of fishing in this change, obviously. It's nigh on impossible to reproduce, or even diagnose, these problems without the physical device on your hand, but this is the best we can do in the circumstances.There are a number of bug reports in Osmdroid repo that give more colour, eg https://github.com/osmdroid/osmdroid/issues/1678On 15 Mar 2024 08:34, Martin ***@***.***> wrote:

which can require a cache reset

Is that a standard Android requirement, documented somewhere?

mvl22:
OK, thanks, that makes sense, having seen that osmdroid ticket.

If this doesn't cause any regressions, I would favour getting this out as a release even if we don't have other things ready - it's clearly an issue affecting multiple people.

jezhiggins:
OK, I'll push that through this evening.

PR #555, opened 2024-04-27
Changed POI id to String and removed hashCode.

### LINKED ISSUE #554: POI ID changed to string
Creating new issue for this.

HilaryN:
To fix #554 
This works, but I'm not sure it's a robust solution as I removed the hashCode override.
@mvl22  is the number within the POI ID unique?  Or perhaps unique within category?

### DIFF
--- a/cyclestreets.app/src/main/assets/whatsnew.html
+++ b/cyclestreets.app/src/main/assets/whatsnew.html
@@ -8,6 +8,7 @@
 <h3>Release 3.12</h3>
 <i>There are no new features in this release. We are updated a number of our
 dependencies, including our underlying maps display.</i>
+<li>POI issue fixed.</li>
 
 <h3>Release 3.11</h3>
 <strong>Features</strong>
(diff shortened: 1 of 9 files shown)


HilaryN:
I've added the hashCode on the String, but not sure if this will be unique?

mvl22:
No, the ID string as a whole is unique. I would simply use that as a unique value if you are having to index the points for some reason.

mvl22:
@oliverlockwood @jezhiggins Could one of you review this so we can push this to beta or straight to live ideally? This continues to generate feedback and it’s clearly a priority for users.

oliverlockwood:
This looks good in principle.

Before merging, I think we should:
- on a local copy of the `master` branch, try **temporarily** removing the `@Ignore` annotation at the top of the `RetrofitApiClientIntegrationTest` and running the two POI-related UTs, namely `hitGetPOIsByBboxApi()` and `hitGetPOIsByRadiusApi()`.  Based on my understanding of the issue, this should **fail**.
- do the same test against the branch for this PR.  This time these two tests should **pass**.
- updating `pois.json` such that at least some of the POIs have alphabetical characters in their ID, not just numerical categories (I have just pushed a commit to your PR that does this)
- this should make the `RetrofitApiClientTest` fail on `master` but pass on your branch.

oliverlockwood (approved):
I've now done the testing actions myself and am satisfied.

HilaryN:
Thank you!

PR #567, opened 2025-04-13
General maintanence updating - photo picker, permissions, build tooling

jezhiggins:
It's been a few months. Just trying to catch up with the latest tooling and language changes and what-not.


mvl22:
Thanks Jez :)

